PR #3, opened 2017-01-25
Don't specify default len/cap sizes to make

bradleyfalzon:
See also original discussion: https://github.com/dominikh/go-simple/pull/29

See testdata for the main tests, but essentially, for any type T:

- `make(T, 0)` can be simplified to `make(T)` for non slice types
- `make(T, 1, 1)` can be simplified to `make(T, 1)`

Notes:
- I'm ignoring the type, as whether the type is a slice, chan or map, the same rules can be applied
- I'm not convinced on the README wording
- `constantInt` is a function copied from `staticcheck`, if this issue is accepted, perhaps it might be better to move this function to `lintuils`?
- I can remove the redundant selector tests, I was just ensuring it was catching the sel.sel.ident case
- Is there a better way to compare idents or selectors? I feel I maybe doing more work than required.
- I haven't ran this through my smaller corpus, I will be doing over the next 48 hours (slowly updating it now), so I haven't checked for false positives, panics etc

### DIFF
--- a/simple/lint.go
+++ b/simple/lint.go
@@ -1385,6 +1385,7 @@ func LintMakeLenCap(f *lint.File) {
 			return true
 		}
 		if fn, ok := call.Fun.(*ast.Ident); !ok || fn.Name != "make" {
+			// FIXME check whether make is indeed the built-in function
 			return true
 		}
 		switch len(call.Args) {
@@ -1393,7 +1394,7 @@ func LintMakeLenCap(f *lint.File) {
 			if _, ok := f.Pkg.TypesInfo.TypeOf(call.Args[0]).Underlying().(*types.Slice); ok {
 				break
 			}
-			if length, ok := call.Args[1].(*ast.BasicLit); ok && length.Value == "0" {
+			if lint.IsZero(call.Args[1]) {
 				f.Errorf(call.Args[1], "when length is zero, length can be omitted")
 			}
 		case 3:
--- a/simple/lint.go
+++ b/simple/lint.go
@@ -34,6 +34,7 @@ var Funcs = map[string]lint.Func{
 	"S1016": LintSimplerStructConversion,
 	"S1017": LintTrim,
 	"S1018": LintLoopSlide,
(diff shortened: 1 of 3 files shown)


simple/testdata/make-lencap.go
```diff
@@ -0,0 +1,17 @@
+package pkg
+
+func fn() {
+	const c = 0
+	var x, y int
+	type s []int
+	_ = make([]int, 1)
+	_ = make([]int, c)       // constant of 0 maybe due to debugging, math or platform-specific code
```
dominikh: Don't use `[]int` to test correct handling of constants, as it will never complain about slices, anyway.

simple/lint.go
```diff
@@ -1280,3 +1281,32 @@ func intLit(f *lint.File, expr ast.Expr) (int64, bool) {
...
 }
+
+func LintMakeLenCap(f *lint.File) {
+	fn := func(node ast.Node) bool {
+		call, ok := node.(*ast.CallExpr)
+		if !ok {
+			return true
+		}
+		if fn, ok := call.Fun.(*ast.Ident); !ok || fn.Name != "make" {
+			return true
```
dominikh: Add a `// FIXME`, saying that we should check whether `make` is indeed the built-in function.

simple/lint.go
```diff
@@ -1280,3 +1281,32 @@ func intLit(f *lint.File, expr ast.Expr) (int64, bool) {
...
+		}
+		switch len(call.Args) {
+		case 2:
+			// make(T, len)
+			if _, ok := f.Pkg.TypesInfo.TypeOf(call.Args[0]).Underlying().(*types.Slice); ok {
+				break
+			}
+			if length, ok := call.Args[1].(*ast.BasicLit); ok && length.Value == "0" {
```
dominikh: You can use `lint.IsZero` here.

bradleyfalzon:
I've made the changes, can you please re-review?

dominikh:
LGTM, will make some slight adjustments (wording) when merging.

PR #32, opened 2017-02-12
Document why you shouldn't call os.Exit in flag.Usage

kevinburke:
I didn't know that the flag package had error handling after Usage() if you
didn't explicitly call it. So I read through the flag source code and found
that it calls the provided ErrorHandling technique after calling Usage().


dominikh:
The README describes some of the checks in more detail, including the flag.Usage check; see https://github.com/dominikh/go-tools/tree/int/cmd/staticcheck#sa1022--calling-osexit-in-a-function-assigned-to-flagusage -- The checks in the table that have further documentation are links, though some of those were broken before. That was fixed in 2a86a384566923fd1db3a0da04afd78a8282e90b.

kevinburke:
Oh okay! I was wondering why clicking the link went nowhere and didn't see the code (SA1022) referenced anywhere else.

dominikh:
Yeah, sorry about the confusion. The tool I use(d) for previewing markdown locally, and GitHub's own engine, generate slightly different anchor tags for headers.

PR #45, opened 2017-02-24
Add -indirect flag to rdeps which prints indirect dependencies.

### LINKED ISSUE #42: rdeps: add a flag to print indirect dependencies
We would like to use `rdeps` for limiting the amount of test one needs to run on pre-push hook. If one modifies a package X, we want to run tests for X and for all packages that depend on X.

Currently, with dependency tree `X <= Y <= Z`, `rdeps X` yields `Y`, and there is no easy way to get both `Y` and `Z`. We would like to implement a `-indirect` (or `-recursive`?) flag, that will result in `rdeps -indirect X` returning both `Y` and `Z`.

I'll happy to implement this and open a pull request, I just want to hear your opinion on that.

mwek:
Closes #42

### DIFF
--- a/cmd/rdeps/rdeps.go
+++ b/cmd/rdeps/rdeps.go
@@ -1,9 +1,8 @@
 // rdeps scans GOPATH for all reverse dependencies of a set of Go
 // packages.
 //
-// rdeps will not sort or deduplicate its output, and the order of the
-// output is undefined. Pipe its output through sort [-u] if you need
-// stable output.
+// rdeps will not sort its output, and the order of the output is
+// undefined. Pipe its output through sort if you need stable output.
 package main
 
 import (
@@ -22,6 +21,7 @@ func main() {
 	var tags buildutil.TagsFlag
 	flag.Var(&tags, "tags", "List of build tags")
 	stdin := flag.Bool("stdin", false, "Read packages from stdin instead of the command line")
+	recursive := flag.Bool("r", false, "Print reverse dependencies recursively")
 	flag.Parse()
 
 	ctx := build.Default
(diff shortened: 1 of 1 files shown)


cmd/rdeps/rdeps.go
```diff
@@ -22,6 +21,7 @@ func main() {
 	var tags buildutil.TagsFlag
 	flag.Var(&tags, "tags", "List of build tags")
 	stdin := flag.Bool("stdin", false, "Read packages from stdin instead of the command line")
+	indirect := flag.Bool("indirect", false, "Print also indirect dependencies")
```
dominikh: Call the flag `r` (for recursive), with the description `Print reverse dependencies recursively`. Call the variable `recursive`.

cmd/rdeps/rdeps.go
```diff
@@ -53,11 +53,24 @@ func main() {
 	}
 	_, reverse, errors := importgraph.Build(&ctx)
 	_ = errors
-	for _, pkg := range pkgs {
+
+	printed := map[string]bool{}
```
dominikh: Call the variable `seen`

cmd/rdeps/rdeps.go
```diff
@@ -53,11 +53,24 @@ func main() {
...
+	var printRDeps func(pkg string)
+	printRDeps = func(pkg string) {
 		for rdep := range reverse[pkg] {
-			fmt.Println(rdep)
+			if !printed[rdep] {
```
dominikh: `if printed[rdep] { continue }`, then you don't need to indent the common case.

mwek:
All comments are addressed, I believe this is ready to be merged.

dominikh:
Merged, thanks.

PR #68, opened 2017-03-16
all: remove a couple of unused func parameters

mvdan:
Found with github.com/mvdan/unparam.


shurcooL:
The tools are collaborating and self-improving. What's next? 😆

mvdan:
They have children, of course.

PR #69, opened 2017-03-16
simple: add rule for unnecessary string copies

### LINKED ISSUE #66: simple: suggest against "copying" of strings
Too often I've seen people doing copies of strings when they never make sense.

Some common examples:

```
string([]byte(s))
fmt.Sprintf("%s", s)
"" + s
```

I'd argue `gosimple` should say that all of those could be just `s`. I can't think of a program that would break because of this.

mvdan:
fmt.Sprintf("%s", s) will be handled by a separate rule.

Fixes #66.

### DIFF
--- a/simple/lint.go
+++ b/simple/lint.go
@@ -56,6 +56,7 @@ func (c *Checker) Funcs() map[string]lint.Func {
 		"S1023": c.LintRedundantBreak,
 		"S1024": c.LintTimeUntil,
 		"S1025": c.LintRedundantSprintf,
+		"S1026": c.LintStringCopy,
 	}
 }
 
@@ -1700,6 +1701,64 @@ func (c *Checker) LintRedundantSprintf(j *lint.Job) {
 			} else {
 				j.Errorf(call, "the argument's underlying type is a string, should use a simple conversion instead of fmt.Sprintf")
 			}
+		}
+		return true
+	}
+	for _, f := range c.filterGenerated(j.Program.Files) {
+		ast.Inspect(f, fn)
+	}
+}
+
+func (c *Checker) LintStringCopy(j *lint.Job) {
+	emptyStringLit := func(e ast.Expr) bool {
+		bl, ok := e.(*ast.BasicLit)
+		return ok && bl.Value == `""`
+	}
+	fn := func(node ast.Node) bool {
+		switch x := node.(type) {
+		case *ast.BinaryExpr: // "" + s, s + ""
+			if x.Op != token.ADD {
+				break
+			}
+			l1 := j.Program.Prog.Fset.Position(x.X.Pos()).Line
+			l2 := j.Program.Prog.Fset.Position(x.Y.Pos()).Line
+			if l1 != l2 {
+				break
+			}
+			var want ast.Expr
+			switch {
+			case emptyStringLit(x.X):
+				want = x.Y
+			case emptyStringLit(x.Y):
+				want = x.X
+			default:
+				return true
+			}
+			j.Errorf(x, "should use %s instead of %s",
+				j.Render(want), j.Render(x))
+		case *ast.CallExpr: // string([]byte(s))
+			bt, ok := j.Program.Info.TypeOf(x.Fun).(*types.Basic)
+			if !ok || bt.Kind() != types.String {
+				break
+			}
+			nested, ok := x.Args[0].(*ast.CallExpr)
+			if !ok {
+				break
+			}
+			st, ok := j.Program.Info.TypeOf(nested.Fun).(*types.Slice)
+			if !ok {
+				break
+			}
+			et, ok := st.Elem().(*types.Basic)
+			if !ok || et.Kind() != types.Byte {
+				break
+			}
+			xt, ok := j.Program.Info.TypeOf(nested.Args[0]).(*types.Basic)
+			if !ok || xt.Kind() != types.String {
+				break
+			}
+			j.Errorf(x, "should use %s instead of %s",
+				j.Render(nested.Args[0]), j.Render(x))
 		}
 		return true
 	}
--- a/simple/testdata/LintStringCopy.go
+++ b/simple/testdata/LintStringCopy.go
@@ -1,0 +1,32 @@
+package pkg
+
+func fn(s string) {
+	_ = string([]byte(s)) // MATCH "should use s instead of string([]byte(s))"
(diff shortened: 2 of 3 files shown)


mvdan:
Open to suggestions on how to improve the actual warning message. Don't know what the policy is regarding messages that include bits of code versus those that don't, e.g. `replace string([]byte(foo)) by foo`.

simple/lint.go
```diff
@@ -1639,3 +1640,47 @@ func (c *Checker) LintRedundantBreak(j *lint.Job) {
...
 	}
 }
+
+func (c *Checker) LintStringCopy(j *lint.Job) {
+	addEmptyStr := func(x, y ast.Expr) bool {
+		bl, ok := x.(*ast.BasicLit)
+		if !ok || bl.Kind != token.STRING || bl.Value != `""` {
+			return false
+		}
+		bt, ok := j.Program.Info.TypeOf(y).(*types.Basic)
```
dominikh: The type check seems counter-productive. Is `x + ""` not a redundant copy if `x` is of `type T string`? The fact that it's a string literal should be enough for soundness of the check. 

If I am mistaken, please add a test that demonstrates the importance of the check.

simple/lint.go
```diff
@@ -1639,3 +1640,47 @@ func (c *Checker) LintRedundantBreak(j *lint.Job) {
...
+		return ok && bt.Kind() == types.String
+	}
+	fn := func(node ast.Node) bool {
+		switch x := node.(type) {
+		case *ast.BinaryExpr: // "" + s, s + ""
+			if addEmptyStr(x.X, x.Y) || addEmptyStr(x.Y, x.X) {
+				j.Errorf(node, "redundant string copy")
```
dominikh: I would avoid all references to the idea of "redundant string copies". Copying the string may or may not be what the user had in mind. I would only point out the facts. something like the following:

`should use x instead of x + ""` and `should use x instead of string([]byte(x))` – use `j.Render` for the specific snippets of code.

simple/testdata/LintStringCopy.go
```diff
@@ -0,0 +1,15 @@
...
+	_ = s + "" // MATCH /redundant string copy/
+
+	_ = s
+	_ = s + "foo"
+
+	string := func(v interface{}) string {
+		return "foo"
+	}
+	_ = string([]byte(s))
+}
```
dominikh: We'll want at least two more tests. One for `string([]rune(s))` and one for `type byte rune; string([]byte(s))`, neither of which should match.

simple/lint.go
mvdan: Ack

simple/testdata/LintStringCopy.go
mvdan: Ack

simple/lint.go
mvdan: You're saying that if `x + ""` type checked properly, `x` must already be of type string so this check is redundant?

dominikh: `x` must be of underlying type `string`, yeah. Nowadays, all input that we work on has been fully type-checked and would compile.

mvdan:
all comments addressed, PTAL

simple/lint.go
```diff
@@ -1707,3 +1708,52 @@ func (c *Checker) LintRedundantSprintf(j *lint.Job) {
...
 }
+
+func (c *Checker) LintStringCopy(j *lint.Job) {
+	emptyStringLit := func(e ast.Expr) bool {
+		bl, ok := e.(*ast.BasicLit)
+		return ok && bl.Value == `""`
+	}
+	fn := func(node ast.Node) bool {
+		switch x := node.(type) {
+		case *ast.BinaryExpr: // "" + s, s + ""
```
dominikh: Check that the operation is `+` and not, for example, `==` – also add a test for that.

mvdan: done

simple/lint.go
```diff
@@ -1707,3 +1708,55 @@ func (c *Checker) LintRedundantSprintf(j *lint.Job) {
...
+			}
+			nested, ok := x.Args[0].(*ast.CallExpr)
+			if !ok {
+				break
+			}
+			st, ok := j.Program.Info.TypeOf(nested.Fun).(*types.Slice)
+			if !ok {
+				break
+			}
+			et, ok := st.Elem().(*types.Basic)
```
dominikh: We need to check that `[]byte(x)` in `string([]byte(x))` is really converting a string to []byte. It could also be something like a json.RawMessage. Taken from the stdlib:

```
if string([]byte(*data.Id)) != raw {
	t.Fatalf("Raw mismatch: have %#q want %#q", []byte(*data.Id), raw)
}
```

where `data.Id` is a *json.RawMessage

simple/lint.go
```diff
@@ -1707,3 +1708,55 @@ func (c *Checker) LintRedundantSprintf(j *lint.Job) {
...
+			if !ok || et.Kind() != types.Byte {
+				break
+			}
+			j.Errorf(x, "should use %s instead of %s",
```
dominikh: If the rendered form of `x` contains newlines, we should not emit the error. In the standard library, I found the following pattern:

```
r := reader("" + // for code formatting purpose.
	"a:\n" +
	" 0 \r\n" +
	"b:1 \t\r\n" +
	"c: 2\r\n" +
	" 3\t\n" +
	"  \t 4  \r\n\n")
```

where the `""` is only used to make the multi-line string easier to read. A image/jpeg test takes it to the extreme with the following:

```
const input = "" +
	"\xff\xd8\xff\xe0\x00\x10\x4a\x46\x49\x46\x00\x01\x01\x00\x00\x01" +
	"\x00\x01\x00\x00\xff\xdb\x00\x43\x00\x10\x0b\x0c\x0e\x0c\x0a\x10" +
	"\x0e\x89\x0e\x12\x11\x10\x13\x18\xff\xd8\xff\xe0\x00\x10\x4a\x46" +
	"\x49\x46\x00\x01\x01\x00\x00\x01\x00\x01\x00\x00\xff\xdb\x00\x43" +
[...]
	"\xf9\x2e\xe0\x0a\x62\x7f\xdf\xd9"
```

It seems likely that the majority of multi-line strings with `""` concatenation in them are for formatting purposes.

mvdan: Good point - will update.

mvdan: Done - `std` only has two matches now, and they seem correct.

PR #85, opened 2017-04-19
simple: add a new check for redundant returns

### LINKED ISSUE #64: simple: suggest dropping unnecessary returns
```
package foo

func Foo() {
        println("bar")
        return
}
```

A return with no args at the end of a func body is unnecessary, as long as the func has no naked return values.

mvdan:
Fixes #64.

I manually checked `gosimple -tests=false std`; it has a handful of warnings and they're all correct.


mvdan:
In action: https://go-review.googlesource.com/c/41093/

dominikh:
Please update the README, rest LGTM.

mvdan:
Ah, forgot about that. Done.

PR #99, opened 2017-05-13
cmd/gosimple/README: add gofmt -r rule for time.Until

shurcooL:
For applying S1024 simplification suggestion.


shurcooL:
FWIW, I tested out this rule (via copy/paste) when making https://github.com/shurcooL/tictactoe/commit/1bb9d2b39738cc527aaf332e77768b6cb93693c1, and that commit passes CI. So I'm pretty confident it's typo-free and accurate.

PR #114, opened 2017-06-01
lint/lintutil: add a flag which allows erroneous input

shkw:
As of now, any tools stop running and give no advice if input contains erroneous packages or files. But it would be really useful if the tools could give any advice for valid packages and files even in such situation.

This PR is to add a new flag which can change the handling of such erroneous input with keeping the current behavior by default.

Actually I still have the concerns below about my changes:
- I am not confident of the flag name `-allow_errors`. It might be better to rename it into more appropriate and simpler one. (e.g. just `-e` like `go list`)
- Tools exit with 0 when there are no problems with valid packages and files even if input contains erroneous packages or files. (should exit with 1 ?)

I would appreciate any feedback and am happy to work further on this.

### DIFF
--- a/lint/lintutil/util.go
+++ b/lint/lintutil/util.go
@@ -117,6 +117,7 @@ func FlagSet(name string) *flag.FlagSet {
 	flags.String("tags", "", "List of `build tags`")
 	flags.String("ignore", "", "Space separated list of checks to ignore, in the following format: 'import/path/file.go:Check1,Check2,...' Both the import path and file name sections support globbing, e.g. 'os/exec/*_test.go'")
 	flags.Bool("tests", true, "Include tests")
+	flags.Bool("allow_errors", false, "Do not stop linting even if erroneous packages or files are found")
 
 	tags := build.Default.ReleaseTags
 	v := tags[len(tags)-1][2:]
@@ -133,13 +134,15 @@ func ProcessFlagSet(c lint.Checker, fs *flag.FlagSet) {
 	tags := fs.Lookup("tags").Value.(flag.Getter).Get().(string)
 	ignore := fs.Lookup("ignore").Value.(flag.Getter).Get().(string)
 	tests := fs.Lookup("tests").Value.(flag.Getter).Get().(bool)
+	allowErrors := fs.Lookup("allow_errors").Value.(flag.Getter).Get().(bool)
 	version := fs.Lookup("go").Value.(flag.Getter).Get().(int)
 
 	ps, lprog, err := Lint(c, fs.Args(), &Options{
-		Tags:      strings.Fields(tags),
-		LintTests: tests,
-		Ignores:   ignore,
-		GoVersion: version,
+		Tags:        strings.Fields(tags),
+		LintTests:   tests,
+		Ignores:     ignore,
+		GoVersion:   version,
+		AllowErrors: allowErrors,
 	})
 	if err != nil {
 		fmt.Fprintln(os.Stderr, err)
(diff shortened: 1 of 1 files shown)


dominikh:
Unfortunately, this isn't really possible. 

- We use go/ssa for building SSA programs. go/ssa requires all packages to be fully type-checked and free of errors. Running it on incomplete packages will lead to panics.
- Our checks all assume that they are operating on code that type checks. We assume that specific function calls have specific number of arguments, that we always have types for expressions, and so on. Ignoring errors will lead to crashing checks.

While we could skip any SSA-based check, and rewrite all the other checks to be robust against incomplete packages, it isn't something I am willing to do. In the case of staticcheck, only a very small subset of checks would even be able to run, and other tools are increasingly relying on SSA.

If you're planning to run the tools on a disjoint set of packages, some of which might be invalid Go, you're better off checking each package separately, in a separate invocation.

s-kawaguchi:
Thank you so much for the prompt reply.

I understood the design of the tools and that my idea does not work well.

> If you're planning to run the tools on a disjoint set of packages, some of which might be invalid Go, you're better off checking each package separately, in a separate invocation.

I have been using some tools in the way like `go list -e ./... | xargs -n1 gosimple` but this takes very long time as I have many (200+) packages to check, so I have also been struggling to improve the performance. Actually this PR was for that purpose.

I have some other ideas I can do to improve the performance without hacking the tools, so I will try them for now. I would appreciate it if anyone could give me some advice in terms of performance improvement on a large code base that might temporarily contain invalid Go code.

Anyway, thank you for providing the useful tools.

shkw:
Thank you so much for the prompt reply.

I understood the design of the tools and that my idea does not work well.

> If you're planning to run the tools on a disjoint set of packages, some of which might be invalid Go, you're better off checking each package separately, in a separate invocation.

I have been using some tools in the way like `go list -e ./... | xargs -n1 gosimple` but this takes very long time as I have many (200+) packages to check, so I have also been struggling to improve the performance. Actually this PR was for that purpose.

I have some other ideas I can do to improve the performance without hacking the tools, so I will try them for now. I would appreciate it if anyone could give me some advice in terms of performance improvement on a large code base that might temporarily contain invalid Go code.

Anyway, thank you for providing the useful tools.

shurcooL:
> I would appreciate it if anyone could give me some advice in terms of performance improvement on a large code base that might temporarily contain invalid Go code.

An idea is to maintain valid Go code at stable commits, and check them out or stash your unfinished changes when you want to do a `megacheck` run.

If you want to test local changes for issues before committing, you can probably get away with running `megacheck` on the modified/affected packages only.

If you have certain packages that are constantly broken and you don't want to include them, then exclude them from being checked.

For example, here are some commands (that I have written down) that I use to test and check all of my supported code:

```
# Test all my (supported) Go packages.
go test -short $(go list github.com/shurcooL/... | grep -v github.com/shurcooL/play/ | grep -v github.com/shurcooL/privateplay/ | grep -v github.com/shurcooL/Conception-go)

# Check all my (supported) Go packages.
megacheck $(go list github.com/shurcooL/... | grep -v github.com/shurcooL/play/ | grep -v github.com/shurcooL/privateplay/ | grep -v github.com/shurcooL/Conception-go)
```

shkw:
Thank you for the advice.

I have not noticed the new tool `megacheck`. It looks really helpful to mitigate the performance issues when running multiple tools. I will give it a try.

> If you have certain packages that are constantly broken and you don't want to include them, then exclude them from being checked.

Actually I have been running the tools as a part of a CI job at every commit regardless of whether it is stable or not in order to find typical issues as soon as possible. And it is uncertain which packages are broken since the code base is actively changing and may be partially broken sometimes by members, so it is difficult to prepare a static list of packages to be excluded.

For better performance, I think I should consider doing just like `megacheck ./...` and expecting the comprehensive results from the tools only when the code base has no invalid Go code.

PR #130, opened 2017-07-07
simple: flag unnecessary conversions when using *bytes.Buffer

### LINKED ISSUE #129: simple: string(buf.Bytes()) -> buf.String() where buf is bytes.Buffer
I've written and committed the following diff:

```diff
 out, err := format.Source(buf.Bytes())
 if err != nil {
-	return err
+	log.Println(err)
+	out = []byte("// gofmt error: " + err.Error() + "\n\n" + string(buf.Bytes()))
 }
```

Luckily, before pushing it, I spotted that `string(buf.Bytes())` could be simplified to `buf.String()`.

I've seen at least one other person do this as well, see https://github.com/neelance/graphql-go/pull/98#discussion_r124367951.

I would imagine this could also apply for `[]byte(buf.String())` -> `buf.Bytes()`.

bradleyfalzon:
Resolves #129 but only works for the test case provided, doesn't work if the selector is map (or other types). For example, the following test will fail:

```go
m := map[string]*bytes.Buffer{"key": buf}
_ = string(m["key"].Bytes())  // MATCH "should use buf.String() instead of string(buf.Bytes())"
_ = []byte(m["key"].String()) // MATCH "should use buf.Bytes() instead of []byte(buf.String())"
```

I had a look to see if this could be simplified further using other linter helpers, but couldn't see anything obvious, let me know if this can be improved.

Other cases in the wild: https://github.com/golang/crypto/blob/a48ac81e47fd6f9ed1258f3b60ae9e75f93cb7ed/openpgp/armor/armor_test.go#L53

Edit: I'm not convinced on the README description and the naming of testdata file or func.

### DIFF
--- a/cmd/gosimple/README.md
+++ b/cmd/gosimple/README.md
@@ -68,6 +68,7 @@ constructs:
 | S1027 | `return` as the final statement of a func body with no return values        | Functions that don't return anything don't need a final return statement |
 | S1028 | `errors.New(fmt.Sprintf(...))`                                              | `fmt.Errorf(...)`                                                        |
 | S1029 | `for _, r := range []rune(s)`                                               | `for _, r := range s`                                                    |
+| S1030 | `[]byte` or `strings` conversion on a *bytes.Buffer `String()` or `Bytes()` | Use `Bytes()` or `String()` methods instead                              |
 
 ## gofmt -r
 
--- a/simple/lint.go
+++ b/simple/lint.go
@@ -67,6 +67,7 @@ func (c *Checker) Funcs() map[string]lint.Func {
 		"S1027": c.LintRedundantReturn,
 		"S1028": c.LintErrorsNewSprintf,
 		"S1029": c.LintRangeStringRunes,
+		"S1030": c.LintBytesConversion,
 	}
 }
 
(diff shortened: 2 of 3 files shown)


bradleyfalzon:
Not even I have more changes to push.

bradleyfalzon:
Resolved the issue of using the wrong selector. My only other concern is if the user aliased the builtin types `byte` or `string` and the best way to handle that. I thought there were tests elsewhere that handled the situation, but I can't find it.

dominikh:
The solution to your "aliasing" (which isn't aliasing) issue is not doing dumb string comparisons. For example, to check if you have a slice of bytes (assuming you already know it's a slice), you could do this: `s.Elem() == types.Universe.Lookup("byte").Type()`.

bradleyfalzon:
> The solution to your "aliasing" (which isn't aliasing) issue is not doing dumb string comparisons.

Just to clarify what is it called? And the string comparison was there until another solution was made apparent, and I wasn't sure if it should be solved, there's plenty of these dumb string comparisons around already.

It looks straight forward, only a few extra lines, did you want it handled?

dominikh:
> Just to clarify what is it called? 

I suppose it'd be shadowing. A package-level identifier called `byte` would shadow the universe-level identifier of the same name.

>  the string comparison was there until another solution was made apparent, and I wasn't sure if it should be solved

FTR, "dumb" wasn't meant in a judging way, but in its neutral meaning of "simple", "basic".

> did you want it handled?

Yes please.

bradleyfalzon:
> FTR, "dumb" wasn't meant in a judging way, but in its neutral meaning of "simple", "basic".

Thanks for the clarification.

I've handled this the way I believe is intended, ready for further review when you/others are, although I didn't include a test for this case, which we may want.

dominikh:
Merged as 5e9ab3832584b6fe46a9a39748c985e835b3764b

Issue #47, opened 2017-03-01
staticcheck: Report incorrect equality checks of time.Time values.


dmitshur:
Package `time` documents that:

> Do not use == with Time values.

(Source: https://godoc.org/time#Time.Equal.)

This probably extends to `!=` for inequality checks too.

shurcooL:
Although https://godoc.org/time#Time says:

> Note that the Go == operator compares not just the time instant but also the Location. Therefore, Time values should not be used as map or database keys without first guaranteeing that the identical Location has been set for all values, which can be achieved through use of the UTC or Local method.

Which kinda implies that there are some legal uses of using `==` operator on `time.Time` values? I'm not sure.

This might also change in Go 1.9 as a result of https://github.com/golang/go/issues/12914.

shurcooL:
This issue might be invalid, or at least headed in that direction in Go 1.9.

From Go master godoc of `time` package:

> Note that the Go == operator includes the monotonic clock reading in its comparison. If time values returned from time.Now and time values constructed by other means (for example, by time.Parse or time.Unix) are meant to compare equal when used as map keys, the times returned by time.Now must have the monotonic clock reading stripped, by setting t = t.AddDate(0, 0, 0). In general, prefer t.Equal(u) to t == u, since t.Equal uses the most accurate comparison available and correctly handles the case when only one of its arguments has a monotonic clock reading.

The phrase "In general, prefer t.Equal(u) to t == u" tells me that it's not illegal to use `==` operator on `time.Time` values, and as a result, might not be a valid suggestion `staticcheck` can make.

dominikh:
Will have to evaluate, putting it on the todo list.

mvdan:
I've used `==` in the past with times I have called `UTC()` on, so I'm inclined to believe this would be too strict.

myitcv:
Just to add a bit of colour to the 👍 I added to @mvdan's comments. We use `time.Time` values as part of map keys which requires us (per the docs) to have guaranteed the locations are identical. As a result, we _can_ use `==` everywhere (and do), implicitly via map key hashing but also explicitly. So having such a rule would be a false positive in our case where, per the docs, there is not an issue of correctness.

mvdan:
To be exact, I meant the same - equality in general, not just `==`. Map keys can't be told to use `.Equal()`.

dominikh:
[...]

PR #134, opened 2017-07-18
SA9004: time equality check

stuartcarnie:
Adds static check for reporting potentially problematic `time.Time` (in)equality comparisons

### DIFF
--- a/cmd/staticcheck/docs/checks/SA9004
+++ b/cmd/staticcheck/docs/checks/SA9004
@@ -1,0 +1,1 @@
+Using == or != operators to compare time.Time is problematic; prefer Time.Equal
--- a/staticcheck/lint.go
+++ b/staticcheck/lint.go
@@ -282,6 +282,7 @@ func (c *Checker) Funcs() map[string]lint.Func {
 		"SA9001": c.CheckDubiousDeferInChannelRangeLoop,
 		"SA9002": c.CheckNonOctalFileMode,
 		"SA9003": c.CheckEmptyBranch,
+		"SA9004": c.CheckTimeEqualityComparison,
 	}
 }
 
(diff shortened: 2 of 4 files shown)


staticcheck/lint.go
```diff
@@ -1187,6 +1188,29 @@ func (c *Checker) CheckDiffSizeComparison(j *lint.Job) {
...
+				if binop.Op != token.EQL && binop.Op != token.NEQ {
+					continue
+				}
+
+				if types.TypeString(binop.X.Type(), nil) != "time.Time" ||
+					types.TypeString(binop.Y.Type(), nil) != "time.Time" {
+					continue
+				}
+
+				j.Errorf(binop, "using == or != operators to compare time is problematic; prefer Time.Equal")
```
dominikh: Reword the message to read `should use time.Time.Equal when comparing time.Time values`

cmd/staticcheck/docs/checks/SA9004
```diff
@@ -0,0 +1 @@
+Using == or != operators to compare time.Time is problematic; prefer Time.Equal
```
dominikh: Wrap operators and identifiers in backticks

dominikh: Say `should use time.Time.Equal instead` instead of `prefer Time.Equal`

dominikh: Needs a paragraph explaining *why* time.Time.Equal should be used instead, but I'll write that myself.

staticcheck/testdata/CheckTimeEqualityComparison.go
```diff
@@ -0,0 +1,27 @@
...
+)
+
+func fn1() {
+	var t1, t2 time.Time
+
+	_ = t1 == t2 // MATCH /using == or != operators to compare time is problematic/
+	_ = t1 != t2 // MATCH /using == or != operators to compare time is problematic/
+}
+
+func fn2() {
```
dominikh: Drop the tests involving pointers. People shouldn't be passing around pointers to times like this.

shurcooL:
Is this check viable? Will it skip cases where `==` operator must be used, e.g., when `time.Time` is a map key?

See #47 for previous discussion of this check (before Go 1.9 changes were finalized).

dominikh:
> Is this check viable?

No, it is not. Thanks for linking to prior discussion that I was utterly incapable of finding myself.

stuartcarnie:
Good discussion – is this PR considered closed, @dominikh?

dominikh:
I'm afraid so.

Issue #67, opened 2017-03-16
simple: spot unnecessary nil checks around a range


mvdan:
```
if strs != nil {
        for _, s := range strs {
                println(s)
        }
}
```

can be

```
for _, s := range strs {
        println(s)
}
```

This is a common misconception among people coming from lanugages like Java, where iterating over a nil list is a crash and not zero work.

mvdan:
I just realised the same case applies for maps. And even when reading a single element:

```
if m != nil {
    v := m[k]
    [...]
}
```

The nil check is redundant there too.

dominikh:
Implemented the check for check + loop in #148. 

Will not address the map indexing case for now. We'd have to make sure that the map isn't also being inserted into, and I'm not yet convinced nil check + map index is seen in the wild. Feel free to open a separate issue if this is a code smell existing in the real world.

PR #148, opened 2017-08-01
simple: adds S1031 LintNilCheckAroundRange

scriptnull:
Fixes https://github.com/dominikh/go-tools/issues/67

I am relatively new to manipulating AST and this code base. So, please take a look at the code and let me know if I have made some naive mistakes.

Thank you.

### DIFF
--- a/simple/lint.go
+++ b/simple/lint.go
@@ -84,20 +84,25 @@ func (c *Checker) LintNilCheckAroundRange(j *lint.Job) {
 			return true
 		}
 
-		if ifCond.Op == token.NEQ && j.IsNil(ifCond.Y) && len(ifSt.Body.List) == 1 {
-			rangeSt, ok := ifSt.Body.List[0].(*ast.RangeStmt)
-			if !ok {
-				return true
-			}
-			ifXIdent, ok := ifCond.X.(*ast.Ident)
-			if !ok {
-				return true
-			}
-			rangeXIdent, ok := rangeSt.X.(*ast.Ident)
-			if !ok {
-				return true
-			}
-			if ifXIdent.Name == rangeXIdent.Name {
+		if ifCond.Op != token.NEQ || !j.IsNil(ifCond.Y) || len(ifSt.Body.List) != 1 {
+			return true
+		}
+
+		rangeSt, ok := ifSt.Body.List[0].(*ast.RangeStmt)
+		if !ok {
+			return true
+		}
+		ifXIdent, ok := ifCond.X.(*ast.Ident)
+		if !ok {
+			return true
+		}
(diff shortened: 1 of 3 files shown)


simple/lint.go
```diff
@@ -68,6 +68,43 @@ func (c *Checker) Funcs() map[string]lint.Func {
 		"S1028": c.LintErrorsNewSprintf,
 		"S1029": c.LintRangeStringRunes,
 		"S1030": c.LintBytesBufferConversions,
+		"S1031": c.LintNilCheckAroundRange,
+	}
+}
+
+func (c *Checker) LintNilCheckAroundRange(j *lint.Job) {
```
dominikh: This check is missing one additional condition: We can only suggest the simplification if ranging over a slice.

In the case of channels, ranging over a nil channel would block forever, which is why the check is needed.

In the case of maps, the loop body may try to insert values in the map. Trying to do that on a nil map panics.

simple/testdata/LintNilCheckAroundRange.go
```diff
@@ -0,0 +1,11 @@
+package pkg
+
+func fn() {
+	str := []string{}
+
+	if str != nil { // MATCH /unnecessary nil check around range/
```
dominikh: Add a couple more test cases, namely:

- a range with no check around it (shouldn't match)
- a check with more statements than just the range (shouldn't match)
- ranging over channels (shouldn't match)
- ranging over maps (shouldn't match)

simple/lint.go
dominikh: Forget my silly comment regarding nil maps. Ranging over a nil map will never execute the loop body. So, we should check that the type is not a channel, slices and maps are fine.

simple/testdata/LintNilCheckAroundRange.go
dominikh: Based on my comment above, ranging over maps **should** match.

scriptnull:
@dominikh Thanks for the review.

I think, we need to also match for nil check around range over strings

```go
	var emptyStr string
	if emptyStr != "" { 
		for _, x := range emptyStr {
			fmt.Println(x)
		}
	}
```
is same as
```go
	var emptyStr string
	for _, x := range emptyStr {
		fmt.Println(x)
	}
```

can I add this case also in this linting rule itself ?

dominikh:
I think the string case is contrived and not worth adding.

scriptnull:
Alright, I will not implement that case then.

scriptnull:
@dominikh I have updated the code. Let me know if you need something more. Thanks.

dominikh:
Merged as 7c33446 and 2ab10e4, thank you.

PR #161, opened 2017-08-18
Log error when TypeOf on composite literal returns nil

GeorgeMac:
This avoids a panic, in favor of logging an error when TypeOf
returns nil due to the type being implied from the slice type
the composite literal is defined within.

Steps to reproduce:

```go
package main

import "fmt"

type Thing struct {
	Stuff string
}

func main() {
	things := []Thing{
		{
			"",
		},
	}

	fmt.Println(things)
}
```

Run keyify on the implied composite literal inside the []Thing.

### DIFF
--- a/cmd/keyify/keyify.go
+++ b/cmd/keyify/keyify.go
@@ -128,7 +128,13 @@ func main() {
 		printComplit(complit, lit, lprog.Fset, lprog.Fset)
 		return
 	}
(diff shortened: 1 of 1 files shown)


dominikh:
Thank you, but we can actually support this case.

GeorgeMac:
cool 👌 even better

PR #195, opened 2017-10-04
simple: simplify sort.Sort(sort.StringSlice(/*[]string*/)) and friends (closes #194)

### LINKED ISSUE #194: gosimple: simplify sort.Sort(sort.StringSlice(a)) and sort.Sort(sort.IntSlice(a))
`sort.Sort(sort.StringSlice(a))` -> `sort.Strings(a)` if a is []string
`sort.Sort(sort.IntSlice(a))` -> `sort.Ints(a)` if a is []int

opennota:
Closes #194.

### DIFF
--- a/simple/lint.go
+++ b/simple/lint.go
@@ -70,6 +70,7 @@ func (c *Checker) Funcs() map[string]lint.Func {
 		"S1029": c.LintRangeStringRunes,
 		"S1030": c.LintBytesBufferConversions,
 		"S1031": c.LintNilCheckAroundRange,
+		"S1032": c.LintSort,
 	}
 }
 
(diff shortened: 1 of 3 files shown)


opennota:
@dominikh , could you review the PR?

dominikh:
I will review the PR in a couple days. Bear with me.

simple/testdata/sort.go
```diff
@@ -0,0 +1,14 @@
...
+func fn() {
+	var a []int
+	sort.Sort(sort.IntSlice(a)) // MATCH "sort.Ints"
+
+	var b []float64
+	sort.Sort(sort.Float64Slice(b)) // MATCH "sort.Float64s"
+
+	var c []string
+	sort.Sort(sort.StringSlice(c)) // MATCH "sort.Strings"
+}
```
dominikh: Test cases for checks should include examples of code that mustn't be flagged. I'm thinking of the following ones:

1. sort.Sort with an arbitrary argument that is not sort.IntSlice/sort.Float64Slice/sort.StringSlice

simple/lint.go
```diff
@@ -1839,3 +1840,55 @@ func (c *Checker) LintNilCheckAroundRange(j *lint.Job) {
...
 	}
 }
+
+func (c *Checker) LintSort(j *lint.Job) {
+	fn := func(node ast.Node) bool {
+		if !j.IsCallToAST(node, "sort.Sort") {
+			return true
+		}
+		call := node.(*ast.CallExpr)
+		if len(call.Args) == 0 {
```
dominikh: Impossible branch. We know that the `sort.Sort` function has exactly one argument.

simple/lint.go
```diff
@@ -1839,3 +1840,55 @@ func (c *Checker) LintNilCheckAroundRange(j *lint.Job) {
...
+
+		typeconv, ok := call.Args[0].(*ast.CallExpr)
+		if !ok {
+			return true
+		}
+		if len(typeconv.Args) == 0 {
+			return true
+		}
+
+		sel, ok := typeconv.Fun.(*ast.SelectorExpr)
```
dominikh: Instead of all of this code, I suggest using `lint.IsCallToAnyAST` and to change that function to return the function name that matched. Then this code simply becomes

```
name, ok := j.IsCallToAnyAST(typeconv, "sort.IntSlice", "sort.Float64Slice", "sort.StringSlice")
if !ok { return true }
```

simple/lint.go
opennota: A code might not compile and nevertheless be syntactically valid. Do we need gosimple to panic on `sort.Sort()`?
`sort` also can be not the stdlib's sort package, but an alias for another package.

dominikh: 1. gosimple (as well as staticcheck and my other linters) only operate on code that type checks, i.e. compiles. Ergo, `sort.Sort()` will never reach us.
2. The `sort` in `sort.Sort` is the absolute, resolved import path of the package. It can neither be an alias for another package, nor a vendored, modified copy of `sort`.

simple/testdata/sort.go
dominikh: Another good test case to add, just to be safe: `sort.Sort(x)` where x is a named type with underlying type `[]int`

simple/lint.go
opennota: I couldn't change IsCallToAnyAST to check this case. It's meant for function calls, not type conversions. Even if I change that:

https://github.com/dominikh/go-tools/blob/4cd09d945fb3beeb0a478f1aa6522e1f03da362c/lint/lint.go#L608-L609

to a type switch, `types.TypeName` still doesn't have a method named `FullName`, and `Name` returns just `IntSlice` etc., not `sort.IntSlice`.

dominikh: OK

opennota:
Updated.

dominikh:
Looks good at a glance. I'll merge it in time for the 2017.3 release. I will have to run it against real-world code to see if it's too noisy, and if we'll realistically need to handle a case like the following:

```
sort.Sort(xs)
sort.Sort(ys)
sort.Sort(sort.IntSlice(zs))
```

which is more consistent than using `sort.Ints`.

dominikh:
Merged as 6217a4b and refined in follow-up commits.

PR #268, opened 2018-02-28
Replaced import paths by Cabify's fork ones

colega:


fortytw2:
I don't think you intended to make this PR to this repo, @colega 🙃

colega:
Obviously that was not intended 😂 
Always the same problem with github's UI :(

PR #270, opened 2018-02-28
added -single-file argument that allows checkers to only check/report…

jb0n:
… on a single file.

I did this primarily so gosimple would be able to integrate with w0rp/ale (it wants to run linters on a single file).  I didn't have much hope for staticcheck/megacheck because on a codebase I work in they literally take hours to run on my machine. I was pleasantly surprised that they can now run on a typical file in ~15s.

### DIFF
--- a/lint/lint.go
+++ b/lint/lint.go
@@ -159,6 +159,7 @@ type Linter struct {
 	Ignores       []Ignore
 	GoVersion     int
 	ReturnIgnored bool
+	SingleFile    string
 
 	automaticIgnores []Ignore
 }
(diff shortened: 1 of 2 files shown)


dominikh:
Thank you, but this is not a mode of operation that I want to support. My tools mirror the CLI functionality of the `go` tool, which operates on import paths or lists of files that specify entire packages. I don't want to add a magic "detect the package that this file is part of" mode. More specifically, I don't want to encourage people to think in terms of individual files. The smallest unit we work on are packages.

I'm aware that w0rp/ale has had problems with my tools, but that's on them. Other tools such as flycheck learned how to work with Go packages, and evidently ale already figured it out for `go build`.

As for your performance statistic: If the tools can run in 15 seconds on "one file", I reckon they won't take much longer on the whole package. You already have to parse and type-check the entire package that the file is contained in, and that's the most expensive step. Running the checks itself is comparatively cheap, so skipping over the functions in the other files shouldn't save much time.

jb0n:
Cool, I kind of figured this was the case. I'll just keep my fork in my repo.

PR #272, opened 2018-03-05
keyify: support build tags

arp242:
Otherwise it won't work on e.g.:

    // +build testdb

    package a

    import "fmt"

    type Foo struct {
        F string
    }

    func main() {
        fmt.Println(Foo{"xx"})
    }

### DIFF
--- a/cmd/keyify/keyify.go
+++ b/cmd/keyify/keyify.go
@@ -15,7 +15,6 @@ import (
 	"log"
 	"os"
 	"path/filepath"
-	"strings"
 
 	"honnef.co/go/tools/version"
 
@@ -30,7 +29,6 @@ var (
 	fJSON      bool
 	fMinify    bool
 	fModified  bool
-	fTags      string
 	fVersion   bool
 )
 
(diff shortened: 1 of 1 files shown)


shurcooL:
It would be better to use `buildutil.TagsFlag` for this. See [here](https://github.com/dominikh/go-tools/blob/bf3d85daf373f9ad9fc6da6da7c010f9ee3e8e0c/cmd/rdeps/rdeps.go#L23) or [here](https://github.com/mdempsky/unconvert/pull/27/files).

Carpetsmoker:
Ah, neat. Didn'y know you had a helper for this.

dmitshur:
Friendly ping @dominikh. What do you think of this change?

dominikh:
My thoughts are surprisingly non-trivial because of go/packages stuff. In short, we still need to port keyify to go/packages to support modules, which previously wasn't doable because of lacking support for modified files. By the time we port to go/packages, this PR will no longer be the correct fix. At this point, the PR really just exists to be a constant reminder for myself to finally fix keyify.

dominikh:
At this point the way forward is likely to abandon the `keyify` utility altogether and to incorporate the functionality with gopls instead.

Closing the PR to release it from limbo.

PR #286, opened 2018-05-27
Fix linting issues

mastertinner:
Simplifies code with `gofmt -s` and fixes two typos


dominikh:
Hi, thanks for your contribution. I have some feedback:

- None of the files in `testdata/` directories should be simplified. These are test cases, and if you ran `go test`, you'd notice that the changes actually break some of the tests. Other tests are "only" rendered non-functional by the changes.
- None of the files in `callgraph/` or `ssa/` should be simplified. These packages are forks of packages from golang.org/x/tools, and we want to keep the differences to a minimum to make merging of upstream changes easier. That does include the order of imports.
- https://github.com/dominikh/go-tools/pull/286/files#diff-c03d29f6e3409ccbd46a9b1a6c05b808L2455 removes the space before the tab. Don't do that. The space belongs to the start of the comment, while the tab belongs to the indentation of the commented code. Same thing a couple lines down.

The remaining changes are fine and I'd be happy to merge them once you've undone the other changes.

mastertinner:
Thanks for the detailed feedback, @dominikh. I reverted the changed you mentioned.

mastertinner:
@dominikh, this should now be ready to get merged. Right?

dominikh:
Yes, everything's looking good. I'll merge your PR soon(tm)

PR #294, opened 2018-07-07
lint/lintutil: use runtime.GOARCH to determine type sizes

kortschak:
This arose while porting @cznic's crt to linux_arm. See https://github.com/cznic/crt/pull/3

Please take a look.


dominikh:
Done in d5cd21c, thanks.

PR #307, opened 2018-08-03
fix paths honnef.co -> github.com

soniah:
fix paths honnef.co -> github.com

I'm behind a corporate firewall, so I needed to fix up the path. Thanks for
your work ;-)


dominikh:
That's unfortunately not a change I can merge. The tools' canonical import path is honnef.co/go/... for a reason. Not only would changing it now break existing users, we'd also forego the advantages of vanity import paths, such as independence from GitHub.

I'm afraid you'll have to maintain your own fork.

PR #313, opened 2018-08-18
Not warn if backquote is included in the regexp string

nasa9084:
When the string contains some backquotes, we cannot use raw string literal.
So S1007 should not be appeared when the string contains backquotes.


dominikh:
Thank you. I'll merge the PR in the next couple of days.

zx8:
@dominikh Ping

dominikh:
The PR has been merged into the `next` branch. `next` will be merged into `master` soon(tm) (likely around New Year's)

PR #351, opened 2018-09-26
staticcheck: new check to detect unreachable case clauses in type switches

ksurent:
Detect type switches like this one:

```
switch v.(type) {
case io.Reader:
    // ...
case io.ReadCloser:
    // unreachable
}
```

and similar.

### DIFF
--- a/staticcheck/lint.go
+++ b/staticcheck/lint.go
@@ -294,6 +294,7 @@ func (c *Checker) Funcs() map[string]lint.Func {
 		"SA9002": c.CheckNonOctalFileMode,
 		"SA9003": c.CheckEmptyBranch,
 		"SA9004": c.CheckMissingEnumTypesInDeclaration,
+		"SA9005": c.CheckUnreachableTypeCases,
 	}
 }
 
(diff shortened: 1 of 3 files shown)


staticcheck/lint.go
```diff
@@ -2788,3 +2789,87 @@ func (c *Checker) CheckMissingEnumTypesInDeclaration(j *lint.Job) {
 		ast.Inspect(f, fn)
 	}
 }
+
+func (c *Checker) CheckUnreachableTypeCases(j *lint.Job) {
+	// Check if T subsumes V in a type switch:
+	// T and V are structurally identical interfaces;
+	// interface T is a superset of V;
```
dominikh: T is a subset of V, not a superset.

staticcheck/lint.go
```diff
@@ -2788,3 +2789,87 @@ func (c *Checker) CheckMissingEnumTypesInDeclaration(j *lint.Job) {
...
+			if !ok {
+				continue
+			}
+
+			// Exclude the 'default' case.
+			if len(cc.List) == 0 {
+				continue
+			}
+
+			Ts := []types.Type{}
```
dominikh: Similarly we can trivially preallocate this slice

staticcheck/lint.go
```diff
@@ -2788,3 +2789,87 @@ func (c *Checker) CheckMissingEnumTypesInDeclaration(j *lint.Job) {
...
+			for _, expr := range cc.List {
+				if T := TypeOf(j, expr); T != nil {
+					Ts = append(Ts, T)
+				}
+			}
+
+			if len(Ts) > 0 {
```
dominikh: If TypeOf won't return nil, then this check is unnecessary.

staticcheck/lint.go
```diff
@@ -2788,3 +2789,87 @@ func (c *Checker) CheckMissingEnumTypesInDeclaration(j *lint.Job) {
...
+	subsumes := func(T, V types.Type) bool {
+		tIface, ok := T.Underlying().(*types.Interface)
+		if !ok {
+			return false
+		}
+
+		return types.Implements(V, tIface)
+	}
+
+	intersect := func(Ts, Vs []types.Type) (types.Type, types.Type, bool) {
```
dominikh: Let's not call this `intersect`; it's not really an intersection. Maybe `subsumesAny`?

staticcheck/lint.go
```diff
@@ -2788,3 +2789,87 @@ func (c *Checker) CheckMissingEnumTypesInDeclaration(j *lint.Job) {
...
+		for i, T := range Ts {
+			for j, V := range Vs {
+				if subsumes(T, V) {
+					return Ts[i], Vs[j], true
```
dominikh: `return T, V, true`

staticcheck/lint.go
```diff
@@ -2788,3 +2789,87 @@ func (c *Checker) CheckMissingEnumTypesInDeclaration(j *lint.Job) {
...
+			}
+
+			// Exclude the 'default' case.
+			if len(cc.List) == 0 {
+				continue
+			}
+
+			Ts := []types.Type{}
+			for _, expr := range cc.List {
+				if T := TypeOf(j, expr); T != nil {
```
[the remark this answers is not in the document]
ksurent: Now that I'm looking at this again, I think not.

My concern was mostly ill–formed and/or partial inputs. Having looked at how the loader and the type checker are initialised, I think we won't ever get in that situation because the check won't even be invoked. Will remove the nil check.

ksurent:
This is ready to be reviewed again.

staticcheck/lint.go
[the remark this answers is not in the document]
dominikh: That's right. We only check packages with well-formed ASTs and no errors during type-checking.

PR #352, opened 2018-09-28
Add -ignore-generated flag to ignore generated files

loderunner:
Similarly to [golint](https://github.com/golang/lint/issues/239), this adds a flag to ignore auto-generated files. Fillowing the [official specification](https://github.com/golang/go/issues/13560#issuecomment-288457920) of generated files, the Linter purposefully ignores any file with a "generated code" comment.

----

One of the modifications implies that a `FileIgnore` with an empty `Checks` always returns `true` upon this file. This might be overstretching though I don't see where I could break.

### DIFF
--- a/lint/lint.go
+++ b/lint/lint.go
@@ -14,6 +14,7 @@ import (
 	"go/token"
 	"go/types"
 	"path/filepath"
+	"regexp"
 	"sort"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


dominikh:
I'll have to decline that contribution. We already have a mechanism for ignoring checks in generated files: each check is responsible for deciding whether it should apply to generated files or not. Stylistic issues and code simplifications do not apply to generated code; bugs, however, should be flagged in generated code. Just because it is generated doesn't mean it is free of bugs, or that bugs are acceptable.

If any check incorrectly flags something in generated code that it shouldn't, then that's a bug that should be fixed in the check.

PR #398, opened 2019-01-22
simple: fix false positives on guarded delete check with else-statements

reillywatson:
This came up when running HEAD against one of my private repos :)


dominikh:
LGTM, will merge tomorrow (I'm in CET and it's close to bed time)

PR #522, opened 2019-06-21
Add a linter to point out redundant calls to http.CanonicalHeaderKey.

### LINKED ISSUE #520: simple: flag the use of net/http.CanonicalHeaderKey when using net/http.Headers methods
When using `Header.Get` or `Header.Set` the keys will be canonicalized automatically. We could flag

```golang
req.Header.Set(http.CanonicalHeaderKey(headerName), headerValue)
req.Header.Get(http.CanonicalHeaderKey(headerName), headerValue)
```

markusthoemmes:
Fixes #520 

Calling `http.CanonicalHeaderKey` on the `key` argument of (http.Header).[Add/Del/Get/Set] is always redundant as these functions canonicalize internally anyway.

**Note:** First time contributor to the project. Feel absolutely free to point out any flaws you see, however minor they might be. Thanks 🙂


dominikh:
Thanks! I'm currently preparing for the 2019.2 release, which means I'll get to this PR soon after that, likely early next week. A cursory read, however, looked good, sans some minor adjustments :-)

simple/lint.go
```diff
@@ -1807,3 +1807,21 @@ func LintSimplifyTypeSwitch(pass *analysis.Pass) (interface{}, error) {
 	pass.ResultOf[inspect.Analyzer].(*inspector.Inspector).Preorder([]ast.Node{(*ast.TypeSwitchStmt)(nil)}, fn)
 	return nil, nil
 }
+
+func LintRedundantCanonicalHeaderKey(pass *analysis.Pass) (interface{}, error) {
+	fn := func(node ast.Node) {
+		call := node.(*ast.CallExpr)
+		for _, function := range []string{"Add", "Del", "Get", "Set"} {
+			fqdn := fmt.Sprintf("(net/http.Header).%s", function)
```
dominikh: Instead of looping over a slice and constructing the fqdn over and over, let's just write the `if` by hand:

```
if !IsCallToAST(pass, call, "(net/http.Header).Add") &&
    !IsCallToAST(pass, call, "(net/http.Header).Del") &&
    ...
```

This has the benefit of making it clear that only up to one element of the slice can ever match. That is, if we were to keep using the loop, we'd want to return after ReportNodeFG, and we'd want to return, instead of continue, when the first argument isn't a call to CanonicalHeaderKey. Using the expanded boolean expression avoids all that.

simple/lint.go
```diff
@@ -1807,3 +1807,21 @@ func LintSimplifyTypeSwitch(pass *analysis.Pass) (interface{}, error) {
...
+			if !IsCallToAST(pass, call, fqdn) {
+				continue
+			}
+			if !IsCallToAST(pass, call.Args[Arg(fqdn+".key")], "net/http.CanonicalHeaderKey") {
+				continue
+			}
+			ReportNodefFG(pass, call, "calling net/http.CanonicalHeaderKey on the key of %s is redundant", fqdn)
```
dominikh: Change `on the key of` to `on the 'key' argument of`

simple/testdata/src/LintRedundantCanonicalHeaderKey/LintRedundantCanonicalHeaderKey.go
```diff
@@ -0,0 +1,16 @@
+package pkg
+
+import "net/http"
+
+func fn1() {
```
dominikh: Just to be thorough, add test cases for the following non-matches:

- using http.CanonicalHeaderKey on the value instead of the key
- using something like `fn(http.CanonicalHeaderKey(...))` as the key
- using something like `http.CanonicalHeaderKey(...) + ...` as the key

simple/lint.go
```diff
@@ -1807,3 +1807,21 @@ func LintSimplifyTypeSwitch(pass *analysis.Pass) (interface{}, error) {
...
+
+func LintRedundantCanonicalHeaderKey(pass *analysis.Pass) (interface{}, error) {
+	fn := func(node ast.Node) {
+		call := node.(*ast.CallExpr)
+		for _, function := range []string{"Add", "Del", "Get", "Set"} {
+			fqdn := fmt.Sprintf("(net/http.Header).%s", function)
+			if !IsCallToAST(pass, call, fqdn) {
+				continue
+			}
+			if !IsCallToAST(pass, call.Args[Arg(fqdn+".key")], "net/http.CanonicalHeaderKey") {
```
dominikh: I don't think we're getting our money's worth by using `Arg` here. Just use `0`.

arg/arg.go
```diff
@@ -8,6 +8,10 @@ var args = map[string]int{
 	"(*sync.Pool).Put.x":                   0,
 	"(*text/template.Template).Parse.text": 0,
 	"(io.Seeker).Seek.offset":              0,
+	"(net/http.Header).Add.key":            0,
```
dominikh: Remove these, as per comment further down

simple/doc.go
```diff
@@ -422,4 +422,11 @@ After:
 		Title: `Use result of type assertion to simplify cases`,
 		Since: "2019.2",
 	},
+
+	"S1035": &lint.Documentation{
+		Title: `Redundant call to http.CanonicalHeaderKey in http.Header functions`,
+		Text: `The functions on http.Header, namely Add, Del, Get and Set, already
+canonicalize the given header name.`,
+		Since: "2019.2",
```
dominikh: Change this to `"Unreleased"`

markusthoemmes:
@dominikh thanks for the thorough and helpful review! The changes should be in there. I rebased for good measure.

markusthoemmes:
@dominikh Your wishes have been granted 🎉

markusthoemmes:
Can't reproduce these CI failures, anything I need to do?

dominikh:
You can rebase on top of master if you want. The CI failures are flakes due to a misconfiguration, which we fixed in 86b08b5391b71f2d6a05d15701c9726ce96330e9.

markusthoemmes:
Good call, rebasing worked @dominikh.

PR #537, opened 2019-07-15
Emit a warning if the same package is imported under different names.

### LINKED ISSUE #467: staticcheck: flag duplicate imports

markusthoemmes:
Fix #467.

Sorry I didn't wait for you to confirm you need help. Feel free to bin this if there's an implementation around already.

As always: Feel free to point out whichever nits you have. I put this into `stylecheck` because the `.`-imports where there as well and it seemed a sensible place.

We could cut down in overhead if needed by not collecting all imports but by only marking subsequent ones (not the first).

### DIFF
--- a/stylecheck/doc.go
+++ b/stylecheck/doc.go
@@ -153,7 +153,7 @@ bug, we prefer the more idiomatic 'if x == 42'.`,
 	},
 
 	"ST1019": &lint.Documentation{
-		Title: `Don't import the same package under different names`,
+		Title: `Importing the same package multiple times`,
 		Since: "Unreleased",
 	},
 }
--- a/stylecheck/doc.go
+++ b/stylecheck/doc.go
@@ -151,4 +151,9 @@ bug, we prefer the more idiomatic 'if x == 42'.`,
 		Title: `Avoid zero-width and control characters in string literals`,
 		Since: "2019.2",
 	},
+
+	"ST1019": &lint.Documentation{
+		Title: `Don't import the same package under different names`,
+		Since: "Unreleased",
+	},
 }
--- a/stylecheck/lint.go
+++ b/stylecheck/lint.go
@@ -80,6 +80,30 @@ func CheckDotImports(pass *analysis.Pass) (interface{}, error) {
 	return nil, nil
 }
 
+func CheckDuplicatedImports(pass *analysis.Pass) (interface{}, error) {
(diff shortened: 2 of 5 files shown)


stylecheck/lint.go
```diff
@@ -80,6 +80,30 @@ func CheckDotImports(pass *analysis.Pass) (interface{}, error) {
 	return nil, nil
 }
 
+func CheckDuplicatedImports(pass *analysis.Pass) (interface{}, error) {
+	for _, f := range pass.Files {
+		// Collect all imports by their import path
+		imports := make(map[string][]*ast.ImportSpec, len(f.Imports))
+		for _, imp := range f.Imports {
+			if current, ok := imports[imp.Path.Value]; !ok {
```
dominikh: This entire if/else can be replaced with an unconditional `imports[imp.Path.Value] = append(imports[imp.Path.Value], imp)`

stylecheck/doc.go
```diff
@@ -151,4 +151,9 @@ bug, we prefer the more idiomatic 'if x == 42'.`,
 		Title: `Avoid zero-width and control characters in string literals`,
 		Since: "2019.2",
 	},
+
+	"ST1019": &lint.Documentation{
+		Title: `Don't import the same package under different names`,
```
dominikh: Change the check's title to `Importing the same package multiple times`.

stylecheck/lint.go
```diff
@@ -80,6 +80,30 @@ func CheckDotImports(pass *analysis.Pass) (interface{}, error) {
...
+			} else {
+				imports[imp.Path.Value] = append(current, imp)
+			}
+		}
+
+		for _, value := range imports {
+			// If there's more than one import per path, we flag that
+			if len(value) > 1 {
+				for _, imp := range value {
+					ReportNodefFG(pass, imp, "should not import the same package under different names")
```
dominikh: Change the message to `should not import the same package multiple times`

dominikh:
LGTM aside from minor nitpicks; will test and probably merge soon.

markusthoemmes:
@dominikh changes as requested are applied I think.

dominikh:
Thanks. I'll review and merge your changes after I'm done with my current task, probably by the end of this week.

dominikh:
Merged as 2ca8611, 10f9868 and 20c72ce.

PR #543, opened 2019-07-25
staticcheck: support for build tags

marcin-bakowski-intive:
Since version 2019.2 build tags have stopped working in staticcheck


dominikh:
Thank you. I've implemented a slightly different fix in 78a9562.

marcin-bakowski-intive:
No problem. When do you plan to release next version, which includes this change?

dominikh:
The plan is to publish a new point release on Monday, which will include this and some other bug fixes.

marcin-bakowski-intive:
Thanks

PR #555, opened 2019-08-02
facts: allow other fact types when running deprecated analysis

matloob:
When preparing the result of the deprecated analysis, allow fact
types other than IsDeprecated to be present. This will allow
this analysis to be run in an environment where other analyses
have populated facts.


dominikh:
This can be closed due to golang/tools@35ef2682e516ea0dd37d51a73b1e957db246ebbf, right?

PR #579, opened 2019-09-08
all: fix typos in non-vendored code

ainar-g:
Not sure about the phrase in `lint/runner.go` though.


dominikh:
Thanks.

"interna" was supposed to be "internals".

PR #610, opened 2019-10-03
Add Slice type IsPointerLike

vahid-sohrabloo:
Slice is a pointer too


dominikh:
In the context of the analyses that IsPointerLike is used in, slices are **not** pointers. The in-memory representation of a slice is a struct with 3 fields, length, capacity, and a pointer to data. You can read more about that at http://staticcheck.io/docs/checks#SA6002

This is also why the tests fail with your change.

vahid-sohrabloo:
thanks.

PR #724, opened 2020-04-11
simple: Fix suggested fixes for S1004

sourya:
Also update go/tools to latest version, as `analysistest.RunWithSuggestedFixes` is
very recent.

Signed-off-by: Sourya Vatsyayan <sourya@deepsource.io>


sourya:
I have removed the tests for suggested fixes for Windows, because the CR/LF issue was breaking the tests.

The analysistest package's `RunWithSuggestedFixes` does `format.Source(editedFile)`, which removes carriage returns from the edited file (because its doc says that it works like gofmt, and gofmt removes any carriage returns). However, they do not do the same `format.Source` for the `.golden` file used for comparison.

dominikh:
Thanks for the patch, I'll look into it within the next several days.

> However, it does not run format.Source for the .golden file used for comparison. Hence, the carriage returns remain

In that case, wouldn't the golden file already be gofmted by us and not contain any carriage returns in the first place?

dominikh:
Alright, looks like the issue is that git in the CI environment automatically converts files to Windows-style line endings. Once we fix that with a `.gitattributes` file, everything works as expected.

I'll have to read up more on git and line endings. The simplest solution would be to adopt what the Go project does: treat all files as binary, i.e. never change line endings. More complex solutions have rules specific to some file extensions.

sourya:
> In that case, wouldn't the golden file already be gofmted by us and not contain any carriage returns in the first place?

I wasn't able to understand why this was happening.

> Alright, looks like the issue is that git in the CI environment automatically converts files to Windows-style line endings.

Now I know. :)

Thank you for looking into it and fixing it.

dominikh:
Update: I'll merge your PR + my changes on top once https://go-review.googlesource.com/c/tools/+/229257 gets merged.

dominikh:
Never mind, I've merged it ahead of time in a4eb707c4ac4bb176e9241aeb886ea7d135f66a4 – thanks again!

PR #771, opened 2020-06-03
[SA1012] Fix tests for suggested fixes

sourya:
The entire package in `bool-cmp` dir was being loaded twice by `analysistest.Run` -- once while loading the package itself, and again as a test package. This resulted in the fixes running on the package in the same files twice. `bool-cmp_test.go` wasn't serving any purpose in the package, hence this commit removes it.

Also, the golden file for the test of SA1012 was partial. This commit fixes that.

Signed-off-by: Sourya Vatsyayan <sourya@deepsource.io>

### DIFF
--- a/staticcheck/testdata/src/checkStdlibUsageNilContext/checkStdlibUsageNilContext.go.golden
+++ b/staticcheck/testdata/src/checkStdlibUsageNilContext/checkStdlibUsageNilContext.go.golden
@@ -1,3 +1,4 @@
+-- use context.Background --
 package pkg
 
 import "context"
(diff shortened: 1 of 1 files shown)


dominikh:
`bool-cmp_test.go` does serve a purpose. We flag comparisons against booleans in non-test code, but we do permit them in test code. That's what the file is testing. analysistest being confused by multiple packages is really a bug in analysistest that should be fixed, not worked around by removing test cases.

The other fix looks right at a glance, but I'm wondering why the test wasn't failing? Is it because analysistest is currently skipping over syntactically invalid fixes?

sourya:
Whoops! I missed that. I'll add a fix in `analysistest`. As for the second testcase, the PR is here: https://github.com/golang/tools/pull/232. The tests would fail if that branch is used.

PR #783, opened 2020-06-11
staticcheck: flag impossible comparisons for len and cap builtin-s

### LINKED ISSUE #750: staticcheck: check builtin len() < 0 conditionals
This is similar to [a previous request](https://github.com/dominikh/go-tools/issues/609).

For the builtin `len()` this could fail:

`if len(foo) < 0 { // Block will never run }`

ksurent:
Fixes #750.

### DIFF
--- a/staticcheck/lint.go
+++ b/staticcheck/lint.go
@@ -3917,7 +3917,7 @@ var builtinLessThanZeroQ = pattern.MustParse(`
 	(Or
 		(BinaryExpr
 			(BasicLit "INT" "0")
-			"<"
+			">"
 			(CallExpr builtin@(Builtin (Or "len" "cap")) _))
 		(BinaryExpr
 			(CallExpr builtin@(Builtin (Or "len" "cap")) _)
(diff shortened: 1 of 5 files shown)


staticcheck/lint.go
```diff
@@ -3912,3 +3912,30 @@ func CheckStaticBitShift(pass *analysis.Pass) (interface{}, error) {
 
 	return nil, nil
 }
+
+var builtinLessThanZeroQ = pattern.MustParse(`
+	(Or
+		(BinaryExpr
+			(BasicLit "INT" "0")
+			"<"
```
dominikh, line 3920: I believe this should be `>`, not `<` – len and cap can very well be larger than 0. 0 can't be larger than len and cap. The same mistake exists in your unit tests.

ksurent: D'oh! Copy–paste strikes again!
Fixed.

dominikh:
Merged as df09c08ed66e57919560b3a4f897d8f030f58157 with slight modifications. Thanks!

PR #870, opened 2020-11-03
staticcheck: add config for error_funcitons (fmt.Errorf, errors.New)

burik666:
Add a list of error functions (`fmt.Errorf`, ʻerrors.New`) to the config file.
This useful for checking custom error functions/methods.

For example:
`staticcheck.conf`
```
error_functions = ["errors.New", "fmt.Errorf", "myproject/logger.Errorf", "(myproject/logger.Logger).Errorf"]
```

`main.go`
```go
package main

import (
	"myproject/logger"
)

func main() {
	logger.Errorf("Failed to")

	l := logger.New()
	l.Errorf("Failed to")
}
```

`logger/logger.go`
```go
package logger

func Errorf(format string, v ...interface{}) {
	//
}

type Logger struct{}

func (l Logger) Errorf(format string, v ...interface{}) {
	//
}

func New() *Logger {
	return &Logger{}
}
```

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -128,6 +128,9 @@ func (cfg Config) Merge(ocfg Config) Config {
 	if ocfg.HTTPStatusCodeWhitelist != nil {
 		cfg.HTTPStatusCodeWhitelist = mergeLists(cfg.HTTPStatusCodeWhitelist, ocfg.HTTPStatusCodeWhitelist)
 	}
+	if ocfg.ErrorFunctions != nil {
+		cfg.ErrorFunctions = mergeLists(cfg.ErrorFunctions, ocfg.ErrorFunctions)
+	}
 	return cfg
 }
 
@@ -143,6 +146,7 @@ type Config struct {
 	Initialisms             []string `toml:"initialisms"`
 	DotImportWhitelist      []string `toml:"dot_import_whitelist"`
 	HTTPStatusCodeWhitelist []string `toml:"http_status_code_whitelist"`
+	ErrorFunctions          []string `toml:"error_functions"`
 }
 
 func (c Config) String() string {
@@ -151,7 +155,8 @@ func (c Config) String() string {
 	fmt.Fprintf(buf, "Checks: %#v\n", c.Checks)
 	fmt.Fprintf(buf, "Initialisms: %#v\n", c.Initialisms)
 	fmt.Fprintf(buf, "DotImportWhitelist: %#v\n", c.DotImportWhitelist)
-	fmt.Fprintf(buf, "HTTPStatusCodeWhitelist: %#v", c.HTTPStatusCodeWhitelist)
+	fmt.Fprintf(buf, "HTTPStatusCodeWhitelist: %#v\n", c.HTTPStatusCodeWhitelist)
+	fmt.Fprintf(buf, "ErrorFunctions: %#v\n", c.ErrorFunctions)
 
 	return buf.String()
 }
@@ -169,6 +174,7 @@ var DefaultConfig = Config{
 	},
 	DotImportWhitelist:      []string{},
 	HTTPStatusCodeWhitelist: []string{"200", "400", "404", "500"},
+	ErrorFunctions:          []string{"errors.New", "fmt.Errorf"},
 }
 
 const ConfigName = "staticcheck.conf"
@@ -240,6 +246,7 @@ func Load(dir string) (Config, error) {
 	conf.Initialisms = normalizeList(conf.Initialisms)
 	conf.DotImportWhitelist = normalizeList(conf.DotImportWhitelist)
 	conf.HTTPStatusCodeWhitelist = normalizeList(conf.HTTPStatusCodeWhitelist)
+	conf.ErrorFunctions = normalizeList(conf.ErrorFunctions)
 
 	return conf, nil
 }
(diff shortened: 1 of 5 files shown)


dominikh:
Thank you! Please note that it might take me a while to look at this, but I will get to it eventually.

burik666:
@dominikh, ping

go/ir/irutil/util.go
```diff
@@ -75,7 +75,7 @@ func Vararg(x *ir.Slice) ([]ir.Value, bool) {
 
 func CallName(call *ir.CallCommon) string {
 	if call.IsInvoke() {
-		return ""
+		return call.Method.FullName()
```
dominikh, line 78: I don't know if I want to settle on these semantics. Specifically, note this:

> Note: an abstract method may be shared by multiple interfaces due to embedding; Value.Type() provides the specific interface used for this call. 

For 

```
type I1 interface {Foo()}
type I2 interface {I1}
func fn(i I2) {i.Foo()}
```

should CallName return I1.Foo or I2.Foo?

stylecheck/lint.go
```diff
@@ -375,11 +375,20 @@ func CheckErrorStrings(pass *analysis.Pass) (interface{}, error) {
...
-				if !irutil.IsCallToAny(call.Common(), "errors.New", "fmt.Errorf") {
+				if !irutil.IsCallToAny(call.Common(), config.For(pass).ErrorFunctions...) {
 					continue
 				}
 
-				k, ok := call.Common().Args[0].(*ir.Const)
+				var arg ir.Value
+				// If it is a call to a method of a structure,
+				// the first argument is receiver
+				if _, ok := call.Common().Args[0].(*ir.Load); ok {
```
dominikh, line 385: This isn't a reliable way of detecting methods. We should check that this is a "call" mode call, and then check that its Value refers to a method, by checking that `Signature.Recv() != nil`.

stylecheck/lint.go
```diff
@@ -375,11 +375,20 @@ func CheckErrorStrings(pass *analysis.Pass) (interface{}, error) {
...
+					arg = call.Common().Args[1]
```
dominikh, line 386: We can't trust users to only provide functions that have the right number of arguments. We need bounds checks, for both the Args[0] and the Args[1] cases.

stylecheck/lint.go
```diff
@@ -375,11 +375,20 @@ func CheckErrorStrings(pass *analysis.Pass) (interface{}, error) {
...
+				} else {
+					arg = call.Common().Args[0]
```
dominikh, line 388: Should we let users configure which argument contains the error string, instead of assuming that it is the first argument?

config/config.go
```diff
@@ -151,7 +155,8 @@ func (c Config) String() string {
 	fmt.Fprintf(buf, "Checks: %#v\n", c.Checks)
 	fmt.Fprintf(buf, "Initialisms: %#v\n", c.Initialisms)
 	fmt.Fprintf(buf, "DotImportWhitelist: %#v\n", c.DotImportWhitelist)
-	fmt.Fprintf(buf, "HTTPStatusCodeWhitelist: %#v", c.HTTPStatusCodeWhitelist)
+	fmt.Fprintf(buf, "HTTPStatusCodeWhitelist: %#v\n", c.HTTPStatusCodeWhitelist)
+	fmt.Fprintf(buf, "ErrorFunctions: %#v\n", c.ErrorFunctions)
```
dominikh, line 159: Do we care about the final newline?

config/example.conf
```diff
@@ -8,3 +8,4 @@ initialisms = ["ACL", "API", "ASCII", "CPU", "CSS", "DNS",
 	"XSS", "SIP", "RTP"]
 dot_import_whitelist = []
 http_status_code_whitelist = ["200", "400", "404", "500"]
+error_functions = ["errors.New", "fmt.Errorf"]
```
dominikh, line 11: All checks that make use of this new option need tests, making sure to test edge cases.

dominikh (changes requested):

stylecheck/lint.go
burik666: Previously, it is checked that this function has arguments and one of them is `error`.

```
				if !irutil.IsCallToAny(call.Common(), config.For(pass).ErrorFunctions...) {
					continue
				}
```

stylecheck/lint.go
burik666: It's not about function arguments, it's about calling a function or calling a method.

config/config.go
burik666: I do not know.

config/example.conf
burik666: If anything, it doesn't break the old behavior.

what is the best way to make tests for this?

go/ir/irutil/util.go
burik666: You call `I2.Foo ()`, you don't need to know anything about the implementation of `I2`.

stylecheck/lint.go
dominikh: My point is that not every function that creates errors takes the error string as the first argument. Thus, users might not only want to add function names, but also specify the index of the error string.

config/example.conf
dominikh: Check `stylecheck/testdata/src`. Each check has a directory of test cases.

burik666: I will try to write tests for this.

PR #880, opened 2020-11-28
Add staticcheck flag to list all checks

### LINKED ISSUE #860: cmd/staticcheck: provide a way to list all checks
Although staticcheck has information on all the checks, including full explanations, there doesn't appear to be a way to get it to show them. The only way is to go to the website, which might not be available if you're offline, and might not exactly reflect the version of staticcheck that you happen to be using at the time.

I suggest that a flag be added: `staticcheck -list`, which would list all the available checks, each on their own line, with the check name first, followed by the first line of explanation.

For example:

```
$ staticcheck -list
S1000 Use plain channel send or receive instead of single-case select
S1001 Replace for loop with call to copy
...
```

eyechoirs:
Closes #860

### DIFF
--- a/lintcmd/cmd.go
+++ b/lintcmd/cmd.go
@@ -515,6 +515,7 @@ func FlagSet(name string) *flag.FlagSet {
 	flags.Bool("debug.no-compile-errors", false, "Don't print compile errors")
 	flags.String("debug.measure-analyzers", "", "Write analysis measurements to `file`. `file` will be opened for appending if it already exists.")
 	flags.String("debug.trace", "", "Write trace to `file`")
+	flags.Bool("list", false, "List all available checks")
 
 	checks := list{"inherit"}
 	fail := list{"all"}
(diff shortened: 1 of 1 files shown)


dominikh:
Thanks for the PR. I'm currently preparing the next release of Staticcheck, and will review/merge your changes afterwards.

lintcmd/cmd.go
```diff
@@ -515,6 +515,7 @@ func FlagSet(name string) *flag.FlagSet {
 	flags.Bool("debug.no-compile-errors", false, "Don't print compile errors")
 	flags.String("debug.measure-analyzers", "", "Write analysis measurements to `file`. `file` will be opened for appending if it already exists.")
 	flags.String("debug.trace", "", "Write trace to `file`")
+	flags.Bool("list", false, "List all available checks")
```
dominikh, line 518: Let's call this `list-checks` instead

lintcmd/cmd.go
```diff
@@ -619,6 +621,20 @@ func ProcessFlagSet(cs []*analysis.Analyzer, fs *flag.FlagSet) {
...
+	if listChecks {
+		titles := make(map[string]string)
+		analyzerNames := make([]string, len(cs))
+		for i, c := range cs {
+			titles[c.Name] = strings.Split(c.Doc, "\n")[0]
+			analyzerNames[i] = c.Name
+		}
+		sort.Strings(analyzerNames)
+		for _, k := range analyzerNames {
+			fmt.Fprintln(os.Stderr, fmt.Errorf("%s %s", k, titles[k]))
```
dominikh, line 633: `fmt.Fprintf(os.Stderr, "%s %s\n", k, titles[k])`

dominikh (changes requested):

dominikh:
Author has become a ghost, is unlikely to improve the PR. Closing.

Issue #868, opened 2020-11-02
Staticcheck flagging bounding variable of for loop


juicemia:
```
=> staticcheck -version
staticcheck 2020.1.6
```

```
=> staticcheck -debug.version
staticcheck 2020.1.6

Compiled with Go version: go1.14.1
Main module:
        honnef.co/go/tools
Dependencies:
        github.com/BurntSushi/toml@v0.3.1 (sum: h1:WXkYYl6Yr3qBf1K79EBnL4mak0OimBfB0XUf9Vl28OQ=)
        golang.org/x/tools@v0.0.0-20200410194907-79a7a3126eef (sum: h1:RHORRhs540cYZYrzgU2CPUyykkwZM78hGdzocOo9P8A=)
        golang.org/x/xerrors@v0.0.0-20191204190536-9bdfabe68543 (sum: h1:E7g+9GITq07hpfrRu66IVDexMakfv52eLZ2CXBWiKr4=)
```

```
=> go version
go version go1.14.1 darwin/amd64
```

```
=> go env
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/Users/hugot/Library/Caches/go-build"
GOENV="/Users/hugot/Library/Application Support/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOINSECURE=""
GONOPROXY=""
GONOSUMDB=""
GOOS="darwin"
GOPATH="/Users/hugot/go"
GOPRIVATE=""
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/usr/local/go"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/local/go/pkg/tool/darwin_amd64"
GCCGO="gccgo"
AR="ar"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD="/Users/hugot/src/k8s-flex/go.mod"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
[...]

PR #883, opened 2020-12-07
staticcheck/doc.go: SA4008 flags loops with unconditional exits

evanj:
The warning SA4008 "variable in loop condition never changes" is
triggered when a loop has an unconditional exit, such as a break
or panic. Since the warning is logged on the line of the loop
which could be far from the unconditional exit, it some cases at
first glance it appears this warning is wrong. To help, add
documentation showing that this triggers for unconditional exits.
This should help the next person who runs into this.

This was the cause of a "false positive" bug that was closed as
working as intended: https://github.com/dominikh/go-tools/issues/868

### DIFF
--- a/staticcheck/doc.go
+++ b/staticcheck/doc.go
@@ -358,6 +358,20 @@ falsify results.`,
 
 	"SA4008": {
 		Title: `The variable in the loop condition never changes, are you incrementing the wrong variable?`,
+		Text: `This happens if the code changes the wrong variable in a
+loop, such as:
+
+    for i := 0; i < 10; x++ { /* ... loop body ... */ }
+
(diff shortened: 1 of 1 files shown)


evanj:
This is vaguely related; I found a few cases where this check does NOT trigger. Would it be useful to file issues for these? For example:

```go
	for i, x := 0, 0; i < 10; x++ {
		fmt.Printf("i=%d x=%d\n", i, x)
	}
```

dominikh:
> Would it be useful to file issues for these?
Yes.

dominikh:
Closed by 4d6bc73e0ff718f929e07799153fd34debf07105 and d4ffe3a9b01b81aea47fd67ffbaf113f17063da9.

Issue #887, opened 2020-12-15
SA4008: Does not trigger on loops with multiple variable declarations


evanj:
The "variable in loop condition never changes" check does not trigger if the loop contains multiple variable declarations in the init statement. I noticed this while testing this check in order to update the documentation in https://github.com/dominikh/go-tools/pull/883

## does not trigger (false positive)

```go
package main

import "fmt"

func main() {
	for i, x := 0, 0; i < 10; x++ {
		fmt.Printf("i=%d x=%d\n", i, x)
	}
}
```

## does trigger (correct)

```go
package main

import "fmt"

func main() {
	var x int
	for i := 0; i < 10; x++ {
		fmt.Printf("i=%d x=%d\n", i, x)
	}
}
```

## version details

staticcheck build from git master commit 903cdb5ed2f937fbf4004274aa4654317f193bc4 (Tue Dec 15 12:10:46 2020 +0100)

```
~/go-tools/staticcheck (master)$ ./staticcheck -debug.version
staticcheck (no version)

Compiled with Go version: go1.15.6
Main module:
        honnef.co/go/tools
Dependencies:
        github.com/BurntSushi/toml@v0.3.1 (sum: h1:WXkYYl6Yr3qBf1K79EBnL4mak0OimBfB0XUf9Vl28OQ=)
        golang.org/x/mod@v0.2.0 (sum: h1:KU7oHjnv3XNWfa5COkzUifxZmxp1TyI7ImMXqFxLwvQ=)
        golang.org/x/tools@v0.0.0-20200609164405-eb789aa7ce50 (sum: h1:59syOWj4+Fl+op4LL8fX1kO7HmbdEWfxlw4tcGvH+y0=)
        golang.org/x/xerrors@v0.0.0-20191204190536-9bdfabe68543 (sum: h1:E7g+9GITq07hpfrRu66IVDexMakfv52eLZ2CXBWiKr4=)
~/go-tools/staticcheck (master)$ go version
go version go1.15.6 linux/amd64
~/go-tools/staticcheck (master)$ go env
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/ej/.cache/go-build"
GOENV="/home/ej/.config/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOINSECURE=""
GOMODCACHE="/home/ej/gopath/pkg/mod"
GONOPROXY=""
[...]

PR #888, opened 2020-12-15
staticcheck CheckLoopCondition: test multiple variable declarations

evanj:
This is currently not flagged but probably should be. See:

https://github.com/dominikh/go-tools/issues/887

This is a failing test, in case it helps with the above issue. Feel free to close if this is not helpful! Thank you!


dominikh:
Thank you. I am going to close this PR, as it isn't really actionable.

PR #892, opened 2020-12-25
SA4020: fix false positive for nil in a case statement

programmer04:
Consider this program:

```go
package main

func main() {
	var x interface{}
	switch x.(type) {
	case interface{}:
		println("This is interface{}")
	case nil:
		println("This is nil")
	}
}
```
the output of execution is correct:
```
This is nil
```

But running `staticcheck` built from a [current master](https://github.com/dominikh/go-tools/tree/903cdb5ed2f937fbf4004274aa4654317f193bc4) reports false-positive:

```
main.go:8:2: unreachable case clause: interface{} will always match before untyped nil (SA4020)
```

With changes introduced in this PR, `statickcheck` won't complain about such a case.

### DIFF
--- a/staticcheck/lint.go
+++ b/staticcheck/lint.go
@@ -3567,9 +3567,12 @@ func CheckUnreachableTypeCases(pass *analysis.Pass) (interface{}, error) {
 				continue
 			}
 
-			Ts := make([]types.Type, len(cc.List))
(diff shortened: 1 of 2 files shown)


dominikh:
At a glance this looks sensible. I'll test and probably merge it soon. Thanks!

PR #893, opened 2020-12-27
Add support for IncDecStmt in arg overwrite

amit-davidson:
I think it straight forward but consider this example:

```
func retry(fn func() error, retries int) error {
	for {
		if err := fn(); err != nil {
			if retries < 1 {
				return err
			}
			retries--
			continue
		}
		return nil
	}
}
```
The linter doesn't flag `retries--`

### DIFF
--- a/staticcheck/lint.go
+++ b/staticcheck/lint.go
@@ -2061,16 +2061,18 @@ func CheckArgOverwritten(pass *analysis.Pass) (interface{}, error) {
 						if assignment != nil {
 							return false
 						}
-						assign, ok := node.(*ast.AssignStmt)
-						if !ok {
-							return true
-						}
-						for _, lhs := range assign.Lhs {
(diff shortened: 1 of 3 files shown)


amit-davidson:
@dominikh the test fails since the check is skipped due to
```

dominikh:
This check flags arguments that get overwritten before their first use. It doesn't flag arguments merely for being overwritten. That is, your example, even if it used assignment instead of post-decrement, would not get flagged. That is also the purpose of the code in https://github.com/dominikh/go-tools/pull/893#issuecomment-751438142 – if the SSA value representing the argument has non-debug referrers, then the value has been used, which means it hasn't been overwritten before its first use.

Of course, incrementing or decrementing a variable always constitutes a read of said variable. Incrementing an argument, therefore, should never trigger this check.

amit-davidson:
@dominikh I see, thanks for your response :) I will close the PR then

PR #914, opened 2021-01-24
staticcheck: new check to flag use of byte slice in io.WriteString

### LINKED ISSUE #839: simple: flag io.WriteString(w, string(b)) for var b []byte
Via https://github.com/dgryski/semgrep-go/commit/95e7194856633a610e810407e368d2005e571dcc

Could be either a simplification or a performance improvement (then a SA6 check).

proishan11:
Fixes #839

### DIFF
--- a/staticcheck/doc.go
+++ b/staticcheck/doc.go
@@ -981,7 +981,7 @@ https://blog.digitalocean.com/how-to-efficiently-compare-strings-in-go/`,
 
 		string([]byte) allocates a buffer to create the string because of the
 		immutable nature of strings in go`,
-		Since: "2020.2",
+		Since: "Unreleased",
 	},
 
 	"SA9001": {
--- a/staticcheck/doc.go
+++ b/staticcheck/doc.go
@@ -975,6 +975,15 @@ https://blog.digitalocean.com/how-to-efficiently-compare-strings-in-go/`,
 		Severity: lint.SeverityWarning,
 	},
 
+	"SA6006": {
+		Title: `Using io.WriteString to write bytes to a writer w`,
+		Text: `w.Write([]byte) should be used instead of io.writeString
+
+		string([]byte) allocates a buffer to create the string because of the
+		immutable nature of strings in go`,
+		Since: "2020.2",
+	},
+
 	"SA9001": {
 		Title:    `Defers in range loops may not run when you expect them to`,
 		Since:    "2017.1",
--- a/staticcheck/lint.go
+++ b/staticcheck/lint.go
@@ -4662,30 +4662,21 @@ func CheckIneffectiveURLQueryModification(pass *analysis.Pass) (interface{}, err
 func CheckByteSliceInIOWriteString(pass *analysis.Pass) (interface{}, error) {
 	fn := func(node ast.Node) {
 		call := node.(*ast.CallExpr)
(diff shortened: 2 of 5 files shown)


staticcheck/doc.go
```diff
@@ -851,6 +851,15 @@ https://blog.digitalocean.com/how-to-efficiently-compare-strings-in-go/`,
 		Since: "2019.2",
 	},
 
+	"SA6006": {
+		Title: `Using io.WriteString to write bytes to a writer w`,
+		Text: `w.Write([]byte) should be used instead of io.writeString
+
+		string([]byte) allocates a buffer to create the string because of the
+		immutable nature of strings in go`,
+		Since: "2020.2",
```
dominikh, line 860: 2020.2 has already been released and does not include this check. New, unreleased checks should say `Since: "Unreleased"` instead.

staticcheck/lint.go
```diff
@@ -4468,3 +4468,35 @@ func CheckBuiltinZeroComparison(pass *analysis.Pass) (interface{}, error) {
 
 	return nil, nil
 }
+
+func CheckByteSliceInIOWriteString(pass *analysis.Pass) (interface{}, error) {
+	fn := func(node ast.Node) {
+		call := node.(*ast.CallExpr)
+
+		if !code.IsCallToAny(pass, call, "io.WriteString") {
```
dominikh, line 4476: Use `code.IsCallTo` instead, since we're only checking for a single name.

staticcheck/lint.go
```diff
@@ -4468,3 +4468,35 @@ func CheckBuiltinZeroComparison(pass *analysis.Pass) (interface{}, error) {
...
+			return
+		}
+
+		arg := call.Args[1]
+		t := pass.TypesInfo.TypeOf(arg)
+		if t == nil {
```
dominikh, line 4482: `t` cannot be nil here.

staticcheck/lint.go
```diff
@@ -4468,3 +4468,35 @@ func CheckBuiltinZeroComparison(pass *analysis.Pass) (interface{}, error) {
...
+			return
+		}
+
+		if conv, ok := arg.(*ast.CallExpr); ok && isName(pass, conv.Fun, "string") {
+			byteArg := conv.Args[0]
```
dominikh, line 4487: The following 6 lines can be simplified to

```
bt, ok := pass.TypesInfo.TypeOf(conv.Args[0]).Underlying().(*types.Slice)
if !ok {
  return
}
```

staticcheck/testdata/src/CheckByteSliceInIOWriteString/CheckByteSliceInIOWriteString.go
```diff
@@ -0,0 +1,16 @@
+package pkg
```
dominikh, line 1: We should also add tests for the cases that we _don't_ want to flag. WriteString with a string, with a string(b) where b is not a []byte.

We should also add tests for more cases we do want to flag, such as string(b) where b's type is T and `type T []byte`.

dominikh (changes requested):

proishan11:
@dominikh Apologies for the making the changes so late. Please review!

PR #978, opened 2021-05-01
S1011: add for i := range s form

### LINKED ISSUE #881: simple: S1011 doesn't fire if the for i := range s form is used
<p>Synopsis:</p>

<pre><code>package main

import "fmt"

func main() {
&#x09;var a = []int{1, 2, 3}
&#x09;var b = []int{4, 5, 6}
&#x09;var c = []int{7, 8, 9}

&#x09;for _, n := range b { /* Line 10. */
&#x09;&#x09;a = append(a, n)
&#x09;}

&#x09;for i := range c { /* Line 14. */
&#x09;&#x09;n := c[i]
&#x09;&#x09;a = append(a, n)
&#x09;}

&#x09;fmt.Println(a)
}</code></pre>

<pre>$ staticcheck --debug.version</pre>

<pre>staticcheck (devel, v0.0.1-2020.1.0.20201124073330-56b7c78ddcd8)

Compiled with Go version: go1.15.4
Main module:
	honnef.co/go/tools@v0.0.1-2020.1.0.20201124073330-56b7c78ddcd8 (sum: h1:FkcSm40GciJwcT9dunF0X8Zv336LFcXaHt0k97IBG4Q=)
Dependencies:
	github.com/BurntSushi/toml@v0.3.1 (sum: h1:WXkYYl6Yr3qBf1K79EBnL4mak0OimBfB0XUf9Vl28OQ=)
	golang.org/x/mod@v0.3.0 (sum: h1:RM4zey1++hCTbCVQfnWeKs9/IEsaBLA8vTkd0WVtmH4=)
	golang.org/x/tools@v0.0.0-20201201064407-fd09bd90d85c (sum: h1:D/mVYXCk6gUcyr7WuGlAk/ShHqgARUXc2VQxo27Hmws=)
	golang.org/x/xerrors@v0.0.0-20200804184101-5ec99f83aff1 (sum: h1:go1bK/D/BFZV2I8cIQd1NKEZ+0owSTG1fDTci4IqFcE=)</pre>

<pre>$ staticcheck --checks=all ./go/tmp/main.go</pre>

<pre>go/tmp/main.go:10:2: should replace loop with a = append(a, b...) (S1011)

Run 'staticcheck -explain <check>' or visit https://staticcheck.io/docs/checks for documentation on checks.</pre>

<p>Line&#xA0;14 should be marked as&#xA0;well.</p>

<p>(Found&#xA0;by&#xA0;@EugeneOne1.)</p>

matsuyoshi30:
Resolves #881

### DIFF
--- a/simple/lint.go
+++ b/simple/lint.go
@@ -767,12 +767,26 @@ func refersTo(pass *analysis.Pass, expr ast.Expr, ident types.Object) bool {
 }
 
 var checkLoopAppendQ = pattern.MustParse(`
+(Or
 	(RangeStmt
 		(Ident "_")
 		val@(Object _)
 		_
 		x
-		[(AssignStmt [lhs] "=" [(CallExpr (Builtin "append") [lhs val])])]) `)
+		[(AssignStmt [lhs] "=" [(CallExpr (Builtin "append") [lhs val])])])
+	(RangeStmt
+		key
+		nil
+		_
+		x
+		[(AssignStmt [lhs] "=" [(CallExpr (Builtin "append") [lhs (IndexExpr val@(Object _) key)])])])
+	(RangeStmt
+		key
+		nil
+		_
+		x
+		[(AssignStmt val@(Object _) ":=" (IndexExpr x key))
+		(AssignStmt [lhs] "=" [(CallExpr (Builtin "append") [lhs val])])]))`)
 
 func CheckLoopAppend(pass *analysis.Pass) (interface{}, error) {
 	fn := func(node ast.Node) {
--- a/simple/lint.go
+++ b/simple/lint.go
@@ -779,7 +779,7 @@ var checkLoopAppendQ = pattern.MustParse(`
 		nil
 		_
 		x
-		[(AssignStmt [lhs] "=" [(CallExpr (Builtin "append") [lhs (IndexExpr val@(Object _) key)])])])
+		[(AssignStmt [lhs] "=" [(CallExpr (Builtin "append") [lhs (IndexExpr x key)])])])
 	(RangeStmt
 		key
 		nil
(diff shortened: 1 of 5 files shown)


simple/lint.go
```diff
@@ -656,12 +656,26 @@ func refersTo(pass *analysis.Pass, expr ast.Expr, ident types.Object) bool {
...
 		_
 		x
-		[(AssignStmt [lhs] "=" [(CallExpr (Builtin "append") [lhs val])])]) `)
+		[(AssignStmt [lhs] "=" [(CallExpr (Builtin "append") [lhs val])])])
+	(RangeStmt
+		key
+		nil
+		_
+		x
+		[(AssignStmt [lhs] "=" [(CallExpr (Builtin "append") [lhs (IndexExpr val@(Object _) key)])])])
```
dominikh, line 671: Shouldn't this be `(IndexExpr x key)` instead of `(IndexExpr val@(Object _) key)`?

dominikh (changes requested):
If we check for these patterns, then we must make sure that when this happens:

```
for i := range fn() {
  x = append(x, fn()[i])
}
```

`fn` doesn't have side effects. Otherwise, calling it repeatedly and calling it once may have different runtime behaviors. We can use `code.MayHaveSideEffects` to check for that.

simple/lint.go
matsuyoshi30: Thanks for review! I fixed at fd22dfa9520ac69c5615eb29b37beb2879eb55fd.

matsuyoshi30:
```
for i := range fn() {
  x = append(x, fn()[i])
}
```

If `fn` returns an array or slice like above, I think `code.MayHaveSideEffects` always returns true because `fn` does load value from memory or stack/heap allocation.

Am I missing something?

dominikh:
> If fn returns an array or slice like above, I think code.MayHaveSideEffects always returns true because fn does load value from memory or stack/heap allocation.

Even if that is the case, it is the desired behavior. We don't want to make the suggestion when side-effects are involved. `MayHaveSideEffects` also covers things other than function calls.

dominikh:
Thanks for bearing with the slow process. To expedite it somewhat, I've made a number of small changes to your PR instead of going through more iterations of reviews.

PR #1020, opened 2021-05-29
config, doc: update default options.

ldez:
https://github.com/dominikh/go-tools/blob/3d8e6a730b44b8a2f6e399270d4045df3e240476/config/config.go#L159-L172

https://staticcheck.io/docs/options

### DIFF
--- a/doc/options.html
+++ b/doc/options.html
@@ -18,7 +18,7 @@
 </p>
 
 <p>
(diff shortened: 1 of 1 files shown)


dominikh:
Thank you. 

We really need to automate this at some point. There are too many places that contain hand-written extracts of the default configuration.

PR #1103, opened 2021-10-18
Use analysistest.TestData everywhere

tamird:


dominikh:
`trimPosition` in lintcmd/lint_test.go also needs to be updated, or it might break when analysistest.TestData gets overwritten.

tamird:
Anyway, done.

PR #1112, opened 2021-10-29
website: add staticcheck.io website

dominikh:
I'm opening this as a pull request to give people a chance to comment on the new website before it goes live. A preview can be found at https://competent-wing-acf0f7.netlify.app/ – let me know if you find any problems. If there is no feedback, the website will go live some time this weekend, replacing the current website at https://staticcheck.io.

---

This adds a new website for Staticcheck. It uses Hugo[1] and the
Docsy[2] theme and is fully open source, unlike the old website.

The only files omitted from this commit are the list of sponsors and
sponsor logos. These are kept out of tree, to keep the repository
small and to avoid unnecessary commits as sponsors come and go.
The website still builds fine without those files, it will only be
missing the sponsors bar at the bottom, as well as the list of
sponsors on the Sponsors page.

To make generating the check documentation easier, lint.Documentation
has gained two new fields, "Before" and "After". These contain before
and after code samples, which were previously part of the "Text"
field.

The website is its own Go module, to prevent its inclusion in the
honnef.co/go/tools module's ZIP. It's part of this repository,
however, so that changes to Staticcheck and its documentation stay in
sync.

[1]: https://gohugo.io/
[2]: https://www.docsy.dev/


dominikh:
The new website is live.

PR #1132, opened 2021-11-22
WIP docs for new -merge and -matrix flags

dominikh:
Opening a PR to solicit feedback on the first iteration of new documentation.

/cc @ainar-g

### DIFF
--- a/website/content/docs/running-staticcheck/cli/build-tags/index.md
+++ b/website/content/docs/running-staticcheck/cli/build-tags/index.md
@@ -1,0 +1,167 @@
+---
+title: Build tags
+description: How to correctly check code that uses build tags
+draft: true
+---
+
+## Introduction
+
+In Go, files can have build tags, which control when said files will be part of a package.
+For example, two files might contain alternate implementations of the same function, targeting Linux and Windows respectively.
+
+Due to this, a single import path really refers to a collection of packages, with any particular package being chosen by a combination of build tags.
+Even if your code doesn't make use of build tags, any of your transitive dependencies might.
+Therefore, running `staticcheck my/package` really only checks one variant of `my/package`.
+
+For more information on Go's build tags, see [`go help buildconstraint`](https://pkg.go.dev/cmd/go#hdr-Build_constraints).
+
+## Implications
+
+Checking packages using a single set of build tags can lead to both false positives and false negatives.
+The reason for false negatives is straightforward: if some code is being excluded by build tags, then we won't check it.
+False positives can be a bit more involved. Consider the following package, in [txtar format](https://pkg.go.dev/golang.org/x/tools/txtar#hdr-Txtar_format):
+
+```go
+-- api_linux.go --
+package pkg
+
+func Entry() { foo() }
+-- api_windows.go --
+package pkg
+
+func Entry() { bar() }
+-- shared.go --
+package pkg
+
+func foo() {}
+func bar() {}
+```
+
+If we don't check the Windows build, then the function `bar` seems unused.
+Similarly, if we don't check the Linux build, `foo` seems unused.
+
+```terminal
+$ GOOS=linux staticcheck
+shared.go:4:6: func bar is unused (U1000)
+$ GOOS=windows staticcheck
+shared.go:3:6: func foo is unused (U1000)
+```
+
+Only when we check both builds do we see that both functions are in fact used.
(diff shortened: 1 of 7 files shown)


website/content/docs/running-staticcheck/cli/build-tags/index.md
```diff
@@ -0,0 +1,167 @@
...
+
+```terminal
+$ GOOS=linux staticcheck -f binary >file1
+$ GOOS=windows staticcheck -f binary >file2
+$ staticcheck -merge file1 file2
+...
+```
+
+
+Alternatively, if no arguments are passed, `staticcheck` will read from standard input instead.
```
ainar-g, line 118: I'd say it's better to repeat the flag:

 >  […] `staticcheck -merge` will read […]

To avoid any confusion and accidentally making people think that a naked call to `staticcheck` will do here.

website/content/docs/running-staticcheck/cli/build-tags/index.md
```diff
@@ -0,0 +1,167 @@
...
+When using the `-matrix` flag, Staticcheck reads a build matrix from standard input.
+The build matrix uses a line-based format, where each line specifies a build name, environment variables and command-line flags.
+Each line is of the format `<name>: [flags and environment...]`, for example `linux-debug: GOOS=linux -tags=debug`.
+Environment variables and flags get passed to `go` when Staticcheck analyzes code, so you can use all flags that `go` supports, such as `-tags` or `-gcflags`, although few flags other than `-tags` are really useful.
+
+<details>
+<summary>Syntax rules of build matrices</summary>
+Build names may consist of Unicode numbers, Unicode letters, and underscores.
+Environment variable names may consist of a-z, A-Z, 0-9, and underscores.
+Flag names may consist of a-z, 0-9, underscores, and dashes, and they must begin with a dash.
```
ainar-g, line 145: On a few of my projects, we use `--flag` instead of `-flag` for visibility and consistency with some GNU-ish tools, so I think it would be useful to mention explicitly, if the two-dash notation is supported or not.

website/content/docs/running-staticcheck/cli/build-tags/index.md
```diff
@@ -0,0 +1,167 @@
...
+Values that contain spaces must be quoted using double quotes.
+Flags with values must use equal signs. That is, <code>-tags=debug</code> is valid, but <code>-tags debug</code> is not.
```
ainar-g, line 147: This is a pretty important detail, and I'm not sure that it should be hidden under a `<details>` tag.

website/content/docs/running-staticcheck/cli/build-tags/index.md
```diff
@@ -0,0 +1,167 @@
...
+
+```terminal
+$ staticcheck -matrix <<EOF
+windows: GOOS=windows
+linux: GOOS=linux
+appengine: GOOS=linux -tags=appengine
+EOF
+root_windows.go:292:47: syscall.StringToUTF16Ptr has been deprecated since Go 1.1: Use UTF16PtrFromString instead.  [windows] (SA1019)
+verify_test.go:1338:7: const issuerSubjectMatchRoot is unused [appengine,linux,windows] (U1000)
+```
```
ainar-g, line 163: Consider splitting the multiline command and the combined output into two separate monospace blocks to make it easier for people who aren't as experienced with Unix shells to understand where one ends and the other begins.

ainar-g (approved):

website/content/docs/running-staticcheck/cli/build-tags/index.md
dominikh: I'll remove that requirement. Instead, we'll change from `<name>: [environment and flags]` to `<name>: [environment] [flags]` – this will remove most of the restrictions on flags and reduce our parsing of flags to splitting on whitespace.

dominikh:
The new commits should address your concerns. We now do minimal parsing (basically just word-splitting and the bare minimum of quoting; I doubt users will ever need values containing spaces, much less need escaping), which means we no longer need to document the permitted syntax. In fact, we intentionally under-specify the parsing: the user shouldn't have to think about it and should just specify environment variables and flags in a natural manner, and it should "just work".

(We basically implement the format from https://github.com/golang/go/issues/39005)

ainar-g (approved):

dominikh:
Merged as f438cef7f36e682cccd1d5c0b031eee5ff2a4c47

PR #1158, opened 2022-01-04
SA1027: add riscv and riscv64 support to test file

piggynl:
`go test ./...` is failing on `riscv64`. (probably because of missing test file?) Adding `riscv64` to `+build` line will fix.

I have tested this patch on `riscv64` but not on `riscv`. I think this will work on `riscv`.


dominikh:
Thanks.

piggynl:
I found that although [`riscv` is a valid `GOARCH` value](https://github.com/golang/go/blob/1c6426505e640799f2a16d6097eed3f83b372b37/src/go/build/syslist.go#L11), go doesn't support `riscv` as a build target. :cry:

PR #1204, opened 2022-03-18
made stylish output a bit more stylish

mpldr:
This Patchset changes the stylish formatter in the following ways:

- Icon displayed in the summary changes based on results
    - a checkmark if there are no issues (except ignored issues, but I they should not count IMO since they were explicitly ignored)
    - an exclamation mark if there are warnings
    - a cross if there were errors found
        - I didn't use the existing cross, as it would clash visually with the checkmark, and didn't use Emojis because not every system has an Emoji font installed
- Icon is red/yellow/green depending on status
- Categories are coloured based on their severity
    - It was always somewhat hard to parse the list without colour as an optical hint to find the most severe issues

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,6 +7,7 @@ require (
 	golang.org/x/exp/typeparams v0.0.0-20220218215828-6cf2b201936e
 	golang.org/x/sys v0.0.0-20211019181941-9d821ace8654
 	golang.org/x/tools v0.1.8
+	mpldr.codes/ansi v1.5.0
 )
 
 require (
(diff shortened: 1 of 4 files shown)


dominikh:
Sorry, but I'm going to reject this change. The `stylish` check is as stylish as it is going to get. Colors in particular add a lot more complexity than I am willing to accept, namely:

- An additional dependency. Staticcheck intentionally minimizes its number of dependencies, especially on third-parties.
- Extra logic for (not) emitting colors, such as extra environment variables and automatic detection of dumb terminals or output redirection.
- People inevitably not being happy with the colors that were chosen.

Purely subjectively speaking, I think that the color output adds too much visual noise. It's also worth pointing out that our severities only really make sense in an editor context and primarily differentiate between "code that is definitely going to crash" and "code that's wrong in some other way". It's not a useful differentiation on the terminal.

PR #1207, opened 2022-03-19
Do not report duplicate '-' name when `xml:"-"`

isimluk:
Addressing false positive reports like:

   Error: pkg/xsd/attribute.go:18:28: SA5008: invalid XML tag: name "-" conflicts with name "attribute" in *github.com/gocomply/xsd2go/pkg/xsd.Attribute.XMLName (staticcheck)
  	refAttr        *Attribute `xml:"-"`

Relevant excerpt from "encoding/xml" documentation:

    - the XMLName field, described above, is omitted.
    - a field with tag "-" is omitted.

See more at https://pkg.go.dev/encoding/xml#Marshal

### DIFF
--- a/staticcheck/fakexml/typeinfo.go
+++ b/staticcheck/fakexml/typeinfo.go
@@ -204,6 +204,11 @@ func StructFieldInfo(f fakereflect.StructField) (*fieldInfo, error) {
 		return finfo, nil
 	}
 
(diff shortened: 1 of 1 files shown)


dominikh:
Thank you, but I will turn this into an issue. There is a more fundamental bug here that isn't limited to `-`

PR #1212, opened 2022-03-24
config: avoid using deprecated function

kevinburke:
Per staticcheck, DecodeReader is deprecated and should be replaced.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -206,7 +206,7 @@ func parseConfigs(dir string) ([]Config, error) {
 			return nil, err
 		}
 		var cfg Config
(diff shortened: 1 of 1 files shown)


dominikh:
If anyone else is confused by the "per staticcheck" comment and can't actually reproduce it: We're currently using v0.4.1 of the dependency, which does deprecate DecodeReader, but doesn't use the proper comment for that. Later versions do use the right marker and allow Staticcheck to actually catch it. I'll have to assume that OP updated the dependency locally.

dominikh (approved):

PR #1252, opened 2022-04-13
lintcmd: remove redundant code

bxlxx:


dominikh:
How is this code redundant? And why would we remove a test case?

bxlxx:
> How is this code redundant? And why would we remove a test case?

Whether the [line](https://github.com/dominikh/go-tools/blob/d694aadcb1f50c2d8ac0a1dd06217ebb9f654764/lintcmd/config.go#L33) is empty has been judged in the calling function ```parseBuildConfigs```.

dominikh:
But `parseBuildConfig` could be called directly, without going through `parseBuildConfigs`. In fact, that's what the fuzz test is doing, and we need to guard against empty lines or we'll panic.

dominikh:
Thanks, but I think the code is fine as is.

PR #1263, opened 2022-05-02
Website: update docsy theme

deining:
The website uses the [docsy](https://docsy.dev) theme. This PR converts this theme from git submodule to hugo module.
It also brings the theme to latest released version [v0,.2.0](https://github.com/google/docsy/releases/tag/v0.2.0).

### DIFF
--- a/website/i18n/en.toml
+++ b/website/i18n/en.toml
@@ -1,3 +1,0 @@
-# Fix a typo in Docsy (although we don't currently use the print support)
-[print_printable_section]
-other = "This is the multi-page printable view of this section."
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 /dist/20??.?.?/
 /dist/20??.?/
 /internal/cmd/irdump/irdump
+/website/.hugo_build.lock
 /website/public
 /website/resources
 /website/assets/img/sponsors


website/go.mod
```diff
@@ -10,9 +10,13 @@ require (
 )
 
 require (
+	github.com/google/docsy v0.2.0 // indirect
```
dominikh, line 13: What will prevent `go mod tidy` from removing this dependency? It's not used by anything that Go can see.

website/go.mod
```diff
@@ -10,9 +10,13 @@ require (
...
-	golang.org/x/mod v0.6.0-dev.0.20220106191415-9b9b3d81d5e3 // indirect
+	golang.org/x/mod v0.6.0-dev.0.20220419223038-86c51ed26bb4 // indirect
 	golang.org/x/sys v0.0.0-20211019181941-9d821ace8654 // indirect
-	golang.org/x/tools v0.1.11-0.20220316014157-77aa08bb151a // indirect
-	golang.org/x/xerrors v0.0.0-20200804184101-5ec99f83aff1 // indirect
+	golang.org/x/tools v0.1.11-0.20220426200323-dcaea06afc12 // indirect
+)
+
+require (
+	github.com/google/docsy v0.2.0 // indirect
```
dominikh, line 21: Duplicate entry

website/i18n/en.toml
```diff
@@ -1,3 +0,0 @@
-# Fix a typo in Docsy (although we don't currently use the print support)
```
dominikh, line 1: Was this fixed upstream?

website/config.toml
```diff
@@ -1,12 +1,24 @@
 baseURL = "https://staticcheck.io"
 languageCode = 'en-us'
 title = 'Staticcheck'
-theme = 'docsy'
```
dominikh, line 4: Does the theme get used automagically, just because we're importing the module?

dominikh (commented):
Thanks, but I don't think that Hugo modules will work for us. In particular the interaction with `go mod tidy` seems problematic. There is `hugo mod tidy`, but that only considers the modules imported by Hugo, not any of our other Go code in this module.

Hugo modules seem to be rather new, and I'd rather not become an early adopter. Our current system works better for us.

website/go.mod
deining: Thanks for poining that out. This is now fixed.

website/i18n/en.toml
deining: Yes, via this [PR](https://github.com/google/docsy/pull/760).

website/config.toml
deining: No. This section, which was added at the end of `config.toml` imports the modules.

```
[module]
[[module.imports]]
path = "github.com/google/docsy"
[[module.imports]]
path = "github.com/google/docsy/dependencies"
```

**Note**: you could also alter the theme line to: 

```
theme = "github.com/google/docsy"
```

Both are valid (and more or less equivalent) solutions.

website/go.mod
deining: Good point. This is an issue indeed, please note that there is a related open [issue](https://github.com/gohugoio/hugo/issues/6115) in the hugo project that addresses this topic.
Since go 1.14, there is a [-modfile](https://tip.golang.org/doc/go1.14#go-flags) flag that instructs the go command to use an alternate `go.mod` file. Could this be a solution?

dominikh:
Thanks; I'll take another look at this in a couple days.

KarthikNayak:
@deining are you still working on this?

deining:
> @deining are you still working on this?

Please keep that issue open. I hope to find time soon to update the @docsy@ submodule only.

deining:
I updated this PR to just bump the theme submodule to latest docsy version and removed your modifications that are not needed any more after this update.

dominikh:
Merged and published, thanks.

PR #1266, opened 2022-05-07
staticcheck: recognize more options in JSON tags

dsnet:
These are supported by a prototype implementation of json.

### DIFF
--- a/staticcheck/lint.go
+++ b/staticcheck/lint.go
@@ -3894,22 +3894,23 @@ func checkJSONTag(pass *analysis.Pass, field *ast.Field, tag string) {
 	//lint:ignore SA9003 TODO(dh): should we flag empty tags?
 	if len(tag) == 0 {
 	}
+	if i := strings.Index(tag, ",format:"); i >= 0 {
+		tag = tag[:i]
+	}
 	fields := strings.Split(tag, ",")
 	for _, r := range fields[0] {
 		if !unicode.IsLetter(r) && !unicode.IsDigit(r) && !strings.ContainsRune("!#$%&()*+-./:<=>?@[]^_{|}~ ", r) {
 			report.Report(pass, field.Tag, fmt.Sprintf("invalid JSON field name %q", fields[0]))
 		}
 	}
-	var co, cs, ci int
+	options := make(map[string]int)
 	for _, s := range fields[1:] {
 		switch s {
-		case "omitempty":
-			co++
 		case "":
 			// allow stuff like "-,"
 		case "string":
-			cs++
 			// only for string, floating point, integer and bool
+			options[s]++
 			tset := typeutil.NewTypeSet(pass.TypesInfo.TypeOf(field.Type))
 			if len(tset.Terms) == 0 {
 				// TODO(dh): improve message, call out the use of type parameters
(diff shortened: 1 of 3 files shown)


staticcheck/testdata/src/CheckStructTags/CheckStructTags.go
```diff
@@ -14,13 +14,13 @@ type T1 struct {
 	F int        `json:",omitempty,omitempty"` //@ diag(`duplicate JSON option "omitempty"`)
 	G int        `json:",omitempty,string"`
 	H int        `json:",string,omitempty,string"` //@ diag(`duplicate JSON option "string"`)
-	I int        `json:",unknown"`                 //@ diag(`unknown JSON option "unknown"`)
```
dominikh, line 17: Instead of removing the expected diags for `unknown`, use a different, actually unknown option. Here and everywhere else.

staticcheck/lint.go
```diff
@@ -3926,20 +3923,28 @@ func checkJSONTag(pass *analysis.Pass, field *ast.Field, tag string) {
 					}
 				}
 			}
-		case "inline":
-			ci++
+			options[s] = options[s] + 1
```
dominikh, line 3926: You can use `++` with maps. Here and everywhere else.

dominikh, line 3926: In the old code we increment `cs` at the top, so that we increment it even if we hit the `continue`. We should maintain that property,

staticcheck/lint.go
```diff
@@ -3926,20 +3923,28 @@ func checkJSONTag(pass *analysis.Pass, field *ast.Field, tag string) {
...
+		if n > 1 {
+			duplicates = append(duplicates, option)
+		}
 	}
-	if ci > 1 {
-		report.Report(pass, field.Tag, `duplicate JSON option "inline"`)
+	if len(duplicates) > 0 {
+		sort.Strings(duplicates)
+		for _, option := range duplicates {
+			report.Report(pass, field.Tag, `duplicate JSON option `+strconv.Quote(option))
```
dominikh, line 3946: Let's use fmt.Sprintf and %q here, for consistency with the rest of the codebase

dominikh (changes requested):

staticcheck/testdata/src/CheckStructTags/CheckStructTags.go
dsnet: Done.

staticcheck/lint.go
dsnet: Done.

staticcheck/lint.go
dsnet: Done.

staticcheck/lint.go
dsnet: Done.

dsnet:
I made a change to the `format:` flag to simply ignore everything afterwards. This might have some false negatives, but that's okay for now.

staticcheck/lint.go
```diff
@@ -3894,22 +3894,23 @@ func checkJSONTag(pass *analysis.Pass, field *ast.Field, tag string) {
...
+	if i := strings.Index(tag, ",format:"); i >= 0 {
+		tag = tag[:i]
+	}
 	fields := strings.Split(tag, ",")
 	for _, r := range fields[0] {
 		if !unicode.IsLetter(r) && !unicode.IsDigit(r) && !strings.ContainsRune("!#$%&()*+-./:<=>?@[]^_{|}~ ", r) {
 			report.Report(pass, field.Tag, fmt.Sprintf("invalid JSON field name %q", fields[0]))
 		}
 	}
-	var co, cs, ci int
```
dsnet, line 3903: It occurred to me that it may have been written this way for performance. If you'd like me to keep the same pattern, let me know.

dominikh: It was, but with this many new options, the readability of the map wins.

dominikh:
Thanks!

PR #1267, opened 2022-05-08
staticcheck: add docs for SA4008

ericlagergren:
Signed-off-by: Eric Lagergren <eric@ericlagergren.com>

### DIFF
--- a/staticcheck/doc.go
+++ b/staticcheck/doc.go
@@ -443,7 +443,24 @@ falsify results.`,
 	},
 
 	"SA4008": {
-		Title:    `The variable in the loop condition never changes, are you incrementing the wrong variable?`,
+		Title: `The variable in the loop condition never changes, are you incrementing the wrong variable?`,
+		Text: `For example:
+
+	for i := 0; i < 10; j++ { ... }
+
+This may also occur when a loop condition does not have a chance to increment. 
+For example, when a loop body contains an unconditional panic:
+
+	func f() {
+		panic("oops")
+	}
+	func g() {
+		for i := 0; i < 10; i++ {
+			// f unconditionally calls panic, which means "i" is
+			// never incremented.
+			f()
+		}
+	}`,
 		Since:    "2017.1",
 		Severity: lint.SeverityWarning,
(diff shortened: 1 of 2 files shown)


staticcheck/doc.go
```diff
@@ -443,7 +443,22 @@ falsify results.`,
 	},
 
 	"SA4008": {
-		Title:    `The variable in the loop condition never changes, are you incrementing the wrong variable?`,
+		Title: `The variable in the loop condition never changes, are you incrementing the wrong variable?`,
+		Text: `For example:
+
+	for i := 0; i < 10; j++ { ... }
+
+Additionally, this can may also occur when a loop condition does not have a
```
dominikh, line 451: `can may` pick one or the other

staticcheck/doc.go
```diff
@@ -443,7 +443,22 @@ falsify results.`,
...
+chance to increment, like when the loop body contains an unconditional panic:
```
dominikh, line 452: `s/like/for example/`

staticcheck/doc.go
```diff
@@ -443,7 +443,22 @@ falsify results.`,
...
+
+	f := func() {
```
dominikh, line 454: This example is unfortunate, because Staticcheck will not actually flag it. It doesn't look through the variable. For it to work we need a named function; or you could just call `panic` in the loop.

dominikh (changes requested):

staticcheck/doc.go
ericlagergren: did done it

stapelberg:
Thanks for sending this PR! I also encountered an issue recently that left me puzzled. Here’s the reproducer:

```go
package main

func repro() error {
	for i := 0; i < 3; i++ {
		return nil
	}
	return nil
}

func main() {
	repro()
}
```

(The `return` statement results in `i` not being incremented.)

I think your message about unconditionally panicing covers this case, but maybe you could add “unconditionally panicing or returning”?

ericlagergren:
🎉

PR #1273, opened 2022-05-14
Update staticcheck-action version on the docs

andreynering:


dominikh:
Thanks. I've also filed #1274 so that this won't be necessary in the future.

PR #1323, opened 2022-10-04
go-staticcheck is the package Debian users want

andrewpollock:
I was poking around at staticcheck today on Debian, and after some initial confusion, determined that the correct package users need to install is `go-staticcheck`. This PR corrects the documentation.


dominikh:
Looks like this is a fairly new development: https://salsa.debian.org/go-team/packages/golang-honnef-go-tools/-/commit/7e74eb29a6dafdb25ffd0698248b2acb440ad3bc

It's a pleasant surprise, the old name wasn't great for discoverability. Thanks for letting me know.

PR #1340, opened 2022-12-08
Preferring early loop exits to reduce indentation

### LINKED ISSUE #1334: New check: enforce early return in loops
Early returns help to reduce nesting levels. I wonder how much effort is it to implement with staticcheck.io. Examples are:

unwanted:
```go
for _, v := range x {
    if v.Enabled {
        // do stuff
    }
}
```

wanted:
```go
for _, v := range x {
    if !v.Enabled {
        continue
    }
    // do stuff
}
```

param108:
Fixes https://github.com/dominikh/go-tools/issues/1334

The check, 
1. looks for for-loops with single children. 
2. It further checks if the child is an *if* condition. 
3. Finally it checks that the *if* condition has more than 10 children.

### DIFF
--- a/stylecheck/lint.go
+++ b/stylecheck/lint.go
@@ -988,3 +988,30 @@ func CheckExportedVarDocs(pass *analysis.Pass) (interface{}, error) {
 	pass.ResultOf[inspect.Analyzer].(*inspector.Inspector).Nodes([]ast.Node{(*ast.GenDecl)(nil), (*ast.ValueSpec)(nil), (*ast.FuncLit)(nil), (*ast.FuncDecl)(nil)}, fn)
 	return nil, nil
 }
+
+func CheckEarlyLoopReturns(pass *analysis.Pass) (interface{}, error) {
+	fn := func(node ast.Node, push bool) bool {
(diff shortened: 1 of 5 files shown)


stylecheck/lint.go
```diff
@@ -988,3 +988,30 @@ func CheckExportedVarDocs(pass *analysis.Pass) (interface{}, error) {
 	pass.ResultOf[inspect.Analyzer].(*inspector.Inspector).Nodes([]ast.Node{(*ast.GenDecl)(nil), (*ast.ValueSpec)(nil), (*ast.FuncLit)(nil), (*ast.FuncDecl)(nil)}, fn)
 	return nil, nil
 }
+
+func CheckEarlyLoopReturns(pass *analysis.Pass) (interface{}, error) {
+	fn := func(node ast.Node, push bool) bool {
+		switch node := node.(type) {
+		case *ast.ForStmt:
+			if push {
```
nfx, line 996: Can we apply the same style check of reducing indentation to this code? ;)

nfx (changes requested):

Issue #1391, opened 2023-04-06
Flag hex.Encode(b, b)


josharian:
Explained at https://github.com/tailscale/tailscale/commit/7f095617f26bac43d91425935432f16965c04873, along with a sample instance in which it converted a strong hash function into an 8 bit hash function.

PR #1424, opened 2023-07-11
Add SA1031

echlebek:
This commit adds SA1031, a check that detects trivially overlapping slices passed to encoders. Currently, encoding/ascii85, encoding/base32, encoding/base64, and encoding/hex are supported. The check will also check if constant nil is passed to dst, when src is non-nil.

There is currently no Phi or Sigma analysis done. I need to come up with test cases for these situations before adding support for them.

Closes: gh-1391


echlebek:
A note, I decided to avoid flagging when dst is nil. It isn't really related to the intent of the check, an unlikely mistake to make, and trivial to find if it is made. Need to rewrite the commit message in light of that.

staticcheck/doc.go
```diff
@@ -332,6 +332,14 @@ the various parsing and formatting functions in \'strconv\'.`,
 		MergeIf:  lint.MergeIfAny,
 	},
 
+	"SA1031": {
+		Title:    `Overlapping byte slices passed to an encoder`,
+		Text:     `In an encoding function of the form Encode(dst, src), dst and src were found to reference the same memory`,
+		Since:    "2023.2",
```
dominikh, line 338: This should say "Unreleased" until we actually make a new release.

staticcheck/doc.go
```diff
@@ -332,6 +332,14 @@ the various parsing and formatting functions in \'strconv\'.`,
 		MergeIf:  lint.MergeIfAny,
 	},
 
+	"SA1031": {
+		Title:    `Overlapping byte slices passed to an encoder`,
+		Text:     `In an encoding function of the form Encode(dst, src), dst and src were found to reference the same memory`,
```
dominikh, line 337: We should probably explain _why_ that is an issue. It wouldn't be one if for every byte read we only wrote one byte.

staticcheck/lint.go
```diff
@@ -117,6 +117,51 @@ func checkValidHostPort(arg int) CallCheck {
...
 }
 
+func checkNonOverlappingDstSrc(dstArg, srcArg int) CallCheck {
+	return func(call *Call) {
+		dst := call.Args[dstArg]
+		src := call.Args[srcArg]
+		if dst.Value == src.Value {
+			// simple case of f(b, b)
+			// add exception for f(nil, nil)
+			_, dstConst := dst.Value.Value.(*ir.Const)
```
dominikh, line 127: This will need to use irutil.Flatten on the two values, to account for sigma nodes.

staticcheck/lint.go
```diff
@@ -117,6 +117,51 @@ func checkValidHostPort(arg int) CallCheck {
 	}
 }
 
+func checkNonOverlappingDstSrc(dstArg, srcArg int) CallCheck {
+	return func(call *Call) {
```
dominikh, line 121: We have multiple places that check for and ignore `nil`. It'd probably make sense to check if either src or dst are nil and return early.

staticcheck/lint.go
```diff
@@ -117,6 +117,51 @@ func checkValidHostPort(arg int) CallCheck {
...
+		case *ir.Slice:
+			switch srcSlice := irutil.Flatten(src.Value.Value).(type) {
+			case *ir.Const:
+				// the source is nil, and therefore doesn't overlap with dst
+				return
+			case *ir.Phi:
+				// TODO(echlebek: add support for phi and sigma nodes
+				return
+			case *ir.Slice:
+				if dstSlice.X != srcSlice.X {
```
dominikh, line 151: This needs to handle sigma nodes (via ir.Flatten). Consider this:

```
func foo(a *[4]byte) {
	x := a[:]

	if true {
		y := a[:]
		hex.Encode(x, y)
	}
}
```

staticcheck/lint.go
```diff
@@ -117,6 +117,51 @@ func checkValidHostPort(arg int) CallCheck {
...
+					// differing underlying arrays, all is well
+					return
+				}
+				if dstSlice.Low == srcSlice.Low {
```
dominikh, line 155: This needs to handle sigma nodes. Consider this:

```
func foo(a *[4]byte) {
	low := 2
	x := a[low:]

	if true {
		y := a[low:]
		hex.Encode(x, y)
	}
}
```

dominikh (changes requested):

staticcheck/lint.go
```diff
@@ -117,6 +117,41 @@ func checkValidHostPort(arg int) CallCheck {
...
+			// one of the arguments is nil, therefore overlap is not possible
+			return
+		}
+		if dst.Value == src.Value {
+			// simple case of f(b, b)
+			dst.Invalid("overlapping dst and src")
+			return
+		}
+		dstSlice, ok := irutil.Flatten(dst.Value.Value).(*ir.Slice)
+		if !ok {
```
dominikh, line 136: For the future (because it's not applicable here once you fix the panics): you could drop the `, ok` and just let Go panic on the type assertion.

staticcheck/lint.go
```diff
@@ -117,6 +117,41 @@ func checkValidHostPort(arg int) CallCheck {
...
+		if dstConst || srcConst {
+			// one of the arguments is nil, therefore overlap is not possible
+			return
+		}
+		if dst.Value == src.Value {
+			// simple case of f(b, b)
+			dst.Invalid("overlapping dst and src")
+			return
+		}
+		dstSlice, ok := irutil.Flatten(dst.Value.Value).(*ir.Slice)
```
dominikh, line 135: `ir.Slice` is the IR instruction for slicing. Not all slices come into existence by something being sliced. For example, analyzing this code panics:

```
func foo(a, b []byte) {
	hex.Encode(a, b)
}
```

As does

```
func foo(n int) {
	a := make([]byte, n)
	b := make([]byte, n)
	hex.Encode(a, b)
}
```

website/data/checks.json
```diff
@@ -988,6 +988,19 @@
 			"Severity": 1,
 			"MergeIf": 0
 		},
+		"SA1031": {
+			"Title": "Overlapping byte slices passed to an encoder",
+			"Text": "In an encoding function of the form Encode(dst, src), dst and src were found to reference the same memory",
+			"TitleMarkdown": "Overlapping byte slices passed to an encoder",
+			"TextMarkdown": "In an encoding function of the form Encode(dst, src), dst and src were found to reference the same memory",
```
dominikh, line 995: This seems to be out of date with the documentation changes you've made.

staticcheck/doc.go
```diff
@@ -332,6 +332,14 @@ the various parsing and formatting functions in \'strconv\'.`,
 		MergeIf:  lint.MergeIfAny,
 	},
 
+	"SA1031": {
+		Title:    `Overlapping byte slices passed to an encoder`,
+		Text:     `In an encoding function of the form Encode(dst, src), dst and src were found to reference the same memory. This can result in src bytes being overwritten before they are read, when the encoder writes more than one byte per src byte.`,
```
dominikh, line 337: You can use `\'` (backslash followed by single quotation mark) to mark text as code. We translate them to backticks for Markdown.

dominikh (changes requested):

staticcheck/testdata/src/example.com/CheckEncodingAscii85/CheckEncodingAscii85.go
```diff
@@ -0,0 +1,46 @@
...
+	"encoding/ascii85"
+	"log"
+)
+
+func fn() {
+	log.Println(ascii85.Encode(nil, nil))
+	log.Println(ascii85.Encode(make([]byte, 0), nil))
+	sliceA := make([]byte, 8)
+	sliceB := make([]byte, 8)
+	log.Println(ascii85.Encode(sliceA, sliceB))
```
dominikh, line 13: We don't have to use the results of the Encode functions for these tests.

staticcheck/doc.go
```diff
@@ -332,6 +332,14 @@ the various parsing and formatting functions in \'strconv\'.`,
 		MergeIf:  lint.MergeIfAny,
 	},
 
+	"SA1031": {
+		Title:    `Overlapping byte slices passed to an encoder`,
+		Text:     `In an encoding function of the form \'Encode(dst, src)\', \'dst\' and \'src\' were found to reference the same memory. This can result in \'src\' bytes being overwritten before they are read, when the encoder writes more than one byte per \'src\' byte.`,
+		Since:    "Unreleased",
+		Severity: lint.SeverityError,
```
dominikh, line 339: I think this should be a warning, not an error, because the code doesn't panic or return an error.

staticcheck/doc.go
```diff
@@ -332,6 +332,14 @@ the various parsing and formatting functions in \'strconv\'.`,
 		MergeIf:  lint.MergeIfAny,
 	},
 
+	"SA1031": {
+		Title:    `Overlapping byte slices passed to an encoder`,
+		Text:     `In an encoding function of the form \'Encode(dst, src)\', \'dst\' and \'src\' were found to reference the same memory. This can result in \'src\' bytes being overwritten before they are read, when the encoder writes more than one byte per \'src\' byte.`,
```
dominikh, line 337: The text should be wrapped to 72 characters per line.

dominikh (approved):
I'll take care of the final touch-ups on my end.

PR #1430, opened 2023-08-07
Added dependabot

Neo2308:
* Added dependabot to keep go dependencies & actions updated.

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -1,0 +1,16 @@
+version: 2
+updates:
+- package-ecosystem: "gomod"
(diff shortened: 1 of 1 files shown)


dominikh:
No thanks.

Neo2308:
Is there a downside of having dependabot?

dominikh:
We update dependencies consciously, i.e. when they contain changes that matter to us. In particular, the `golang.org/x/tools` module contains a multitude of packages, few of which we actually care about. There's no reason for us to chase updates when they only affect packages we don't use.

PR #1435, opened 2023-08-20
Suggest strings.ContainsFunc instead of strings.IndexFunc with Go 1.21+

scop:


scop:
Modernize/go fix only do slices.Contains(Func), not the corresponding strings/bytes ones staticcheck does.

Issue #1294, opened 2022-06-27
bug: generics: func(X).Y is unused(U1000)


phenpessoa:
Hello!

I've seen two similar issues: https://github.com/dominikh/go-tools/issues/1289 and https://github.com/dominikh/go-tools/issues/1282

But as they are not quite the same as this one, I think I'd open an issue about it, to make sure it is known for the unused rewrite.

Code:
```go
type Foo[T any] interface {
    bar() T
}

type FooBar struct{}

func (f *FooBar) bar() string {
    return "foobar"
}
```
This will yield: `func (*FooBar).bar is unused (U1000)`

The same code without generics won't show any lint warnings, i.e:
```go
type Foo interface {
    bar() string
}

type FooBar struct{}

func (f *FooBar) bar() string {
    return "foobar"
}
```

Staticcheck version: 2022.1.2 (v0.3.2)
Go version:  1.18.3

Rory-Z:
Any update？ I'm using 0.3.3 version in go 1.19.4, and this issue is still exist, but my colleagues don't have this problem and they work well, did I miss something?

Rory-Z:
> No, we didn't fix this yet. I don't know why your colleagues wouldn't have the problem.

Thank‘s for your  answer, I'm not sure why my colleagues wouldn't have the problem, we're working remotely, maybe they got it wrong somewhere, anyway, I've ignored "unused" in GitHub workflow for now, and looking forward to your fix, thank's again.

mvdan:
Raising my hand as well: it's only recently that I've started encountering generics in codebases at work, so I'm starting to see a few unused false positives as well. I _think_ they are all caused by this bug, because the methods being marked as unused then mark other top-level funcs and types as unused as well.

Rory-Z:
> No, we didn't fix this yet. I don't know why your colleagues wouldn't have the problem.

Do we have a roadmap for when to fix it?

dominikh:
> Do we have a roadmap for when to fix it?

I'm working on it.

Rory-Z:
> I'm working on it.

Great, I'm looking forward

qfornaguera:
Hello there! 

I'm dropping here a case I've wrongly posted in golangci-lint repository. Maybe it helps

```
type builder[B any] interface {
	// WithSomething mocks build setting.
	WithSomething() B
	build() struct{}
}

type baseBuilder[B any] struct {
	builder B
}

[...]

Issue #1440, opened 2023-08-24
U1000: A false positive when using generic interfaces


adamroyjones:
This is with staticcheck 2023.1.5 (v0.4.5) and Go 1.21.0.

Here's a very contrived main.go file.

```go
// main.go
package main

import (
	"errors"
	"fmt"
)

func main() {
	f(s{})
}

func f(ii i[error]) {
	fmt.Printf("%v\n", ii.m())
}

type i[T any] interface {
	m() T
}

type s struct{}

func (ss s) m() error {
	return errors.New("from s")
}
```

This runs happily and prints out "from s". However:

```
$ staticcheck ./...
main.go:22:13: func s.m is unused (U1000)
```

If this becomes

```go
package main

import (
	"errors"
	"fmt"
)

func main() {
	f(s{})
}

func f(ii i) {
	fmt.Printf("%v\n", ii.m())
}

type i interface {
	m() error
}

type s struct{}

func (ss s) m() error {
	return errors.New("from s")
}
```

then all is well.

Do you have any thoughts around this? I appreciate that it seems like a goofy example, but it's a reduced version of something a little less goofy.

all2ham:
whats the latest on this

ajscholl:
I think I also ran into this today, for me this program triggered the bug:

```go
package main

import "fmt"

func main() {
	run(myStruct{})
}

func run[E any, T myInterface[E]](repo T) {
	repo.notUnused(nil)
}

type myStruct struct{}

func (myStruct) notUnused(_ []any) {
	fmt.Println("Not unused called")
}

type myInterface[E any] interface {
	notUnused(events []E)
}

// uncommenting removes the lint
// var _ myInterface[any] = myStruct{}
```

PR #1448, opened 2023-09-24
U1000: reduce false positive that relates to type parameters

seiyab:
Addresses https://github.com/dominikh/go-tools/issues/1294.

**NOTE**
Probably some false-positives for complex generic types still exist. And this doesn't address other similar problems such as https://github.com/dominikh/go-tools/issues/1440.

Though I consider this makes a improvement, feel free to reject this if you feel this incomplete fix can be harmful.

### DIFF
--- a/unused/implements.go
+++ b/unused/implements.go
@@ -94,6 +94,7 @@ func newMethodChecker() *methodsChecker {
 	}
 }
 
+// Currently, this doesn't support methods like `foo(x []T)`.
 func (c *methodsChecker) methodIsCompatible(implFunc *types.Func, interfaceFunc *types.Func) bool {
 	if types.Identical(implFunc.Type(), interfaceFunc.Type()) {
 		return true
@@ -149,7 +150,7 @@ func (c *methodsChecker) typeIsCompatible(implType, interfaceType types.Type) bo
 func satisfiesConstraint(t types.Type, tp *types.TypeParam) bool {
 	bound, ok := tp.Underlying().(*types.Interface)
 	if !ok {
-		return false
+		panic("unexpected failure on type assertion (types.Type to *types.Interface)")
 	}
 	return types.Implements(t, bound)
 }
--- a/unused/implements.go
+++ b/unused/implements.go
@@ -1,6 +1,8 @@
 package unused
 
-import "go/types"
+import (
+	"go/types"
+)
 
 // lookupMethod returns the index of and method with matching package and name, or (-1, nil).
 func lookupMethod(T *types.Interface, pkg *types.Package, name string) (int, *types.Func) {
@@ -60,6 +62,7 @@ func implements(V types.Type, T *types.Interface, msV *types.MethodSet) ([]*type
 
 	// A concrete type implements T if it implements all methods of T.
 	var sels []*types.Selection
+	c := newMethodChecker()
 	for i := 0; i < T.NumMethods(); i++ {
 		m := T.Method(i)
 		sel := msV.Lookup(m.Pkg(), m.Name())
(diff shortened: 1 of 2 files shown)


seiyab:
@dominikh any chance to get reviewed?

dominikh:
I'll take a look when I get the chance

dominikh:
I'm wondering if instead we can use https://pkg.go.dev/golang.org/x/exp/typeparams#GenericAssignableTo — that is, replace our implementation of implements with a modified version of GenericAssignableTo.

seiyab:
Thank you for your review. I'll take a look.

seiyab:
I locally tried https://pkg.go.dev/golang.org/x/exp/typeparams#GenericAssignableTo and it successfully passed the tests.
Perhaps it might also resolve opposite false negative as  https://github.com/dominikh/go-tools/issues/1440 reports. I'll investigate more and might add more tests.

seiyab:
@dominikh 
I found that we can't utilize [types.GenericAssignableTo](https://pkg.go.dev/golang.org/x/exp/typeparams#GenericAssignableTo).
types.GenericAssignableTo looks to consider V and T have same numbers of type parameters. (If not, [it falls back types.AssignableTo](https://cs.opensource.google/go/x/exp/+/7918f672:typeparams/common.go;l=150-152))
> If V and T are generic named types, then V is considered assignable to T if, for every possible instantation of V[A_1, ..., A_N], the instantiation T[A_1, ..., A_N] is valid and V[A_1, ..., A_N] implements T[A_1, ..., A_N]. 

In contrast, to resolve https://github.com/dominikh/go-tools/issues/1294, we want to check whether concrete type V is assignable to one of instantiation of T[A_1, ..., A_N] or not.

dominikh (changes requested):
Overall the code looks fine.

I'm very torn on the overall change, however. On the one hand, this fixes some of the common false positives. On the other hand, the incomplete nature of the change means that minor changes to an interface can reintroduce a false positive.

For example, we'd be able to correctly handle `type I[T any] interface { foo() T }`, but not `type I[T any] interface { foo() []T }`. This might be more confusing for users, who will not understand why a minor change to the interface introduces a new false positive.

/cc @mvdan for his opinion.

Note that I don't think we can realistically solve this problem unless we reuse the whole unification step of go/types, which unfortunately isn't currently exported. I'll look into rectifying that.

dominikh:
> @dominikh I found that we can't utilize [types.GenericAssignableTo](https://pkg.go.dev/golang.org/x/exp/typeparams#GenericAssignableTo). types.GenericAssignableTo looks to consider V and T have same numbers of type parameters. (If not, [it falls back types.AssignableTo](https://cs.opensource.google/go/x/exp/+/7918f672:typeparams/common.go;l=150-152))
> 
> > If V and T are generic named types, then V is considered assignable to T if, for every possible instantation of V[A_1, ..., A_N], the instantiation T[A_1, ..., A_N] is valid and V[A_1, ..., A_N] implements T[A_1, ..., A_N].
> 
> In contrast, to resolve #1294, we want to check whether concrete type V is assignable to one of instantiation of T[A_1, ..., A_N] or not.

Thank you for looking into this.

unused/implements.go
```diff
@@ -72,11 +75,81 @@ func implements(V types.Type, T *types.Interface, msV *types.MethodSet) ([]*type
...
+	}
+	return types.Identical(c.typeParams[tp], implType)
+}
+
+func satisfiesConstraint(t types.Type, tp *types.TypeParam) bool {
+	bound, ok := tp.Underlying().(*types.Interface)
+	if !ok {
+		return false
+	}
+	return types.Implements(t, bound)
```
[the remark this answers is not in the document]
seiyab: Thanks, changing it into `types.Satisfies()` make some tests fail and clarify that I'm misunderstanding something around here. Probably `bound` should also be `tp.Constraint()` or something but it sometimes returns `Type` which isn't `*types.Interface` so it can't obviously passed to `types.Satisfies()`.
I'll investigate it and add more test cases.

seiyab:
@dominikh Please take another look.
Thank you for your review. Some my confusions are solved and it get better.

For avoiding allocation, I can't find a lightweight approach to know whether a `*type.Interface` has any type parameters or not. So I use a value of `methodChecker` instead its reference. It can't go out of the function `implements()` scope and should be put on stack. And I also `make()` typeParams map lazily.

mvdan:
@dominikh I think your analysis above is sensible. I think we should push `go/types` to expose more APIs if they are useful for tools.

dominikh:
> @dominikh I think your analysis above is sensible. I think we should push `go/types` to expose more APIs if they are useful for tools.

FWIW, I've filed https://github.com/golang/go/issues/63982

unused/implements.go
```diff
@@ -72,11 +75,85 @@ func implements(V types.Type, T *types.Interface, msV *types.MethodSet) ([]*type
...
 		sels = append(sels, sel)
 	}
 	return sels, true
 }
+
+type methodsChecker struct {
+	typeParams map[*types.TypeParam]types.Type
+}
+
+func newMethodChecker() methodsChecker {
```
dominikh, line 91: We don't need this constructor.

unused/implements.go
```diff
@@ -60,6 +62,7 @@ func implements(V types.Type, T *types.Interface, msV *types.MethodSet) ([]*type
 
 	// A concrete type implements T if it implements all methods of T.
 	var sels []*types.Selection
+	c := newMethodChecker()
```
dominikh, line 65: This can simply be `var c methodsChecker`

unused/implements.go
```diff
@@ -72,11 +75,85 @@ func implements(V types.Type, T *types.Interface, msV *types.MethodSet) ([]*type
...
+		c.typeParams[tp] = implType
+		return true
+	}
+	return types.Identical(c.typeParams[tp], implType)
+}
+
+func satisfiesConstraint(t types.Type, tp *types.TypeParam) bool {
+	bound, ok := tp.Constraint().Underlying().(*types.Interface)
+	if !ok {
+		panic("unexpected failure on type assertion (types.Type to *types.Interface)")
```
dominikh, line 156: Just do `bound := tp.Constraint().Underlying().(*types.Interface)` — we don't need to check for ok just to manually panic — the panic we get on the failed type assertion is fine.

unused/implements.go
```diff
@@ -72,11 +75,81 @@ func implements(V types.Type, T *types.Interface, msV *types.MethodSet) ([]*type
...
+		}
+	}
+	return true
+}
+
+func (c *methodsChecker) typeIsCompatible(implType, interfaceType types.Type) bool {
+	if types.Identical(implType, interfaceType) {
+		return true
+	}
+	tp, ok := interfaceType.(*types.TypeParam)
```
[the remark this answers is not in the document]
dominikh: This still needs a comment stating that we only support trivial uses of type parameters and that this doesn't constitute full unification.

dominikh (changes requested):

seiyab:
@dominikh Thank you for your review. I fixed them.

dominikh (approved):
LGTM, will merge in a couple days.

Issue #1113, opened 2021-10-31
ST1018: doesn't take emoji ZWJ sequences into account (aka "Staticcheck doesn't like [2015+] fun!")


dmitshur:
The latest release version of `staticcheck` reports:

```
string literal contains the Unicode format character U+200D, consider using the '\u200d' escape sequence instead (ST1018)
```

On the following 3 lines:

- https://go.googlesource.com/build/+/refs/changes/97/358897/1/internal/task/tweet.go#297
- https://go.googlesource.com/build/+/refs/changes/97/358897/1/internal/task/tweet.go#320
- https://go.googlesource.com/build/+/refs/changes/97/358897/1/internal/task/tweet.go#321

As far as I can tell it's a false positive, because those 3 strings contain Emoji ZWJ Sequences¹ ² and would not be more readable to write using \u200d in between their composite parts. This is what their unicode sequences are:

"👩🏽‍🔬" aka "👩🏽\u200d🔬": 15 bytes, 4 runes
0: U+1F469 - woman
1: U+1F3FD - skin-tone-4
2: U+200D - Zero Width Joiner (ZWJ)
3: U+1F52C - microscope

"🏃‍♀️" aka "🏃\u200d♀️": 13 bytes, 4 runes
0: U+1F3C3 - runner
1: U+200D - ZWJ
2: U+2640 - male-sign
3: U+FE0F - Variation Selector-16

"🏃‍♂️" aka "🏃\u200d♂️": 13 bytes, 4 runes
0: U+1F3C3 - runner
1: U+200D - ZWJ
2: U+2642 - female-sign
3: U+FE0F - Variation Selector-16

IMO "👩🏽‍🔬", "🏃‍♀️", "🏃‍♂️" are more readable than "👩🏽\u200d🔬", "🏃\u200d♀️", "🏃\u200d♂️".

<details><summary>Screenshot of emoji above</summary>
<img src="https://user-images.githubusercontent.com/1924134/139599530-cb989c92-d0ca-4fdc-ad02-1e9e65641db5.png"></details>

<details><summary>Environment information</summary><br>

```
$ staticcheck -version
staticcheck 2021.1.1 (v0.2.1)

$ staticcheck -debug.version
staticcheck 2021.1.1 (v0.2.1)

Compiled with Go version: go1.17.2
Main module:
[...]

PR #1456, opened 2023-10-19
ST1018: don't flag emojis containing U+FE0F,U+200D

1sixth:
Background: <https://github.com/dominikh/go-tools/issues/1113>

`staticcheck/v0.4.6` reports

```
string literal contains the Unicode format character U+200D, consider using the '\u200d' escape sequence instead (ST1018)
```

for emojis containing the sequence `U+FE0F,U+200D`[^1]. The reason is that `U+FE0F` is not in `unicode.S`. This PR adds a special check for it.

<details>
  <summary>Example Program</summary>

```go
package main

import (
	"fmt"
	"unicode"
)

func main() {
	// https://en.wikipedia.org/wiki/Zero-width_joiner#Examples
	emojis := [...]string{
		"👨‍👩‍👦",
		"🏳️‍🌈",
	}

	for _, emoji := range emojis {
		fmt.Println(emoji)
		for _, e := range emoji {
			fmt.Printf("%+12q unicode.S: %v\n", e, unicode.Is(unicode.S, e))
		}
	}
}
```

</details>

[^1]: There are lots of them, see <https://www.unicode.org/Public/emoji/15.1/emoji-zwj-sequences.txt>

### DIFF
--- a/stylecheck/lint.go
+++ b/stylecheck/lint.go
@@ -732,7 +732,7 @@ func CheckInvisibleCharacters(pass *analysis.Pass) (interface{}, error) {
 			if unicode.Is(unicode.Cf, r) {
 				// Don't flag joined emojis. These are multiple emojis joined with ZWJ, which some platform render as single composite emojis.
 				// For the purpose of this check, we consider all symbols, including all symbol modifiers, emoji.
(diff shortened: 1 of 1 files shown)


dominikh:
Thank you, but I went with a more exhaustive approach.

Issue #1397, opened 2023-04-23
SA4023: staticcheck panics "unhandled type *ir.ArrayConst"


leonidkit:
`staticcheck -version`
`staticcheck 2023.1.3 (v0.4.3)`

`staticcheck -debug.version`
```
Compiled with Go version: go1.20.2
Main module:
        honnef.co/go/tools@v0.4.3 (sum: h1:o/n5/K5gXqk8Gozvs2cnL0F2S1/g1vcGCAx2vETjITw=)
Dependencies:
        github.com/BurntSushi/toml@v1.2.1 (sum: h1:9F2/+DoOYIOksmaJFPw1tGFy1eDnIJXg+UHjuD8lTak=)
        golang.org/x/exp/typeparams@v0.0.0-20221208152030-732eee02a75a (sum: h1:Jw5wfR+h9mnIYH+OtGT2im5wV1YGGDora5vTv/aa5bE=)
        golang.org/x/mod@v0.7.0 (sum: h1:LapD9S96VoQRhi/GrNTqeBJFrUjs5UHCAtTlgwA5oZA=)
        golang.org/x/sys@v0.3.0 (sum: h1:w8ZOecv6NaNa/zC8944JTU3vz4u6Lagfk4RPQxv92NQ=)
        golang.org/x/tools@v0.4.1-0.20221208213631-3f74d914ae6d (sum: h1:9ZNWAi4CYhNv60mXGgAncgq7SGc5qa7C8VZV8Tg7Ggs=)
```

`go version`
`go version go1.20.2 darwin/amd64`

Example:
```
package main

import "github.com/google/uuid"

type ChatID uuid.UUID
type UserID uuid.UUID

type Types interface {
 ChatID | UserID
}

func Parse[T Types](val string) (T, error) {
 parsed, err := uuid.Parse(val)
 if err != nil {
  return T{}, err  // problem line
 }
 return T(parsed), nil
}
```

```
go install honnef.co/go/tools/cmd/staticcheck@latest
```

```
staticcheck ./...
```

```
panic: internal error: unhandled type *ir.ArrayConst

goroutine 319 [running]:
honnef.co/go/tools/analysis/facts/nilness.impl.func1({0x1e90840, 0xc000223c80})
        /Users/llkitaev/go/pkg/mod/honnef.co/go/tools@v0.4.3/analysis/facts/nilness/nilness.go:239 +0x854
[...]

PR #1462, opened 2023-10-31
analysis/facts/nilness: support *ir.ArrayConst

Antonboom:
Closes https://github.com/dominikh/go-tools/issues/1397


staticcheck/testdata/src/example.com/CheckTypedNilInterface/generics.go
```diff
@@ -2,7 +2,7 @@
 
 package pkg
 
-func tpgen1[T *int,]() T {
+func tpgen1[T *int]() T {
```
Antonboom, line 5: `go fmt`, not me

PR #1471, opened 2023-11-13
Add a check for ignored errors in function calls

peterzeller:
Hi there. This is my first PR here, and I did not find a contributing guide. So please guide me through what I need to do :) 

This PR adds a check for unhandled errors. This check is one of the cases where Intellij Goland gives a warning, but staticcheck does not.

The rule here is simple. You get a warning if:

- You have a `CallExpr` as direct parent of an `ExprStmt` (meaning the result is not used)
- And: The called function has an `error` as the last returned type


dominikh:
Thank you for your contribution, but unfortunately this doesn't fit into Staticcheck. Not all unchecked errors are actually problematic, which means these reports would constitute false positives, which we try to avoid as much as possible.

For users who do want this kind of check, there is [errcheck](https://github.com/kisielk/errcheck).

PR #1483, opened 2023-12-15
Add Analyzer alias to sa9003

prestonvanloon:
sa9003 was missing a public `Analyzer` which would conform with the other static checks and work well with tools that expect the package to expose `Analyzer` alias.


dominikh:
Thank you for catching that.

Issue #1251, opened 2022-04-12
staticcheck installation fails with go1.16


elulcao:
Installing staticcheck@latest with go1.16 failed due to `builder.go:36:2: //go:build comment without // +build`

Why the go:build error at line 36, any problem with old Go versions?

https://github.com/dominikh/go-tools/blob/d694aadcb1f50c2d8ac0a1dd06217ebb9f654764/go/ir/builder.go#L36

- The output of 'go version'
```bash
go1.16 version
go version go1.16 darwin/amd64
```

- Exactly which command you ran
- Output of the command and what's wrong with the output
```bash
go1.16 install -v honnef.co/go/tools/cmd/staticcheck@latest
go/pkg/mod/honnef.co/go/tools@v0.3.0/go/ir/builder.go:36:2: //go:build comment without // +build comment
```

- The output of 'go env'
```bash
GO111MODULE="auto"
GOARCH="amd64"
GOBIN=""
GOCACHE="/Users/elulcao/Library/Caches/go-build"
GOENV="/Users/elulcao/Library/Application Support/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOINSECURE=""
GOMODCACHE="/Users/elulcao/go/pkg/mod"
GONOPROXY=""
GONOSUMDB=""
GOOS="darwin"
GOPATH="/Users/elulcao/go"
GOPRIVATE=""
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/Users/elulcao/sdk/go1.16"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/Users/elulcao/sdk/go1.16/pkg/tool/darwin_amd64"
GOVCS=""
GOVERSION="go1.16"
GCCGO="gccgo"
AR="ar"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD=""
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
[...]

PR #1487, opened 2023-12-27
Update staticcheck x GitHub action guide

novalagung:
Changelogs:

- Use the GitHub `actions/checkout@v4` instead of the old one. also, remove the `fetch-depth` property as the default is already set to `1`.
- Fix the go versions in the example. the go `v1.16.x` is not supported by staticcheck `v2022.1.1`
  - use go `v1.21.x` & `v1.22.x` instead
  - reference: https://github.com/dominikh/go-tools/issues/1251#issuecomment-1096744855
- adjust space/indentation

### DIFF
--- a/website/content/docs/running-staticcheck/ci/github-actions/index.md
+++ b/website/content/docs/running-staticcheck/ci/github-actions/index.md
@@ -24,7 +24,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: dominikh/staticcheck-action@v1.3.0
         with:
-          version: "2022.1.1"
+          version: "2023.1.6"
 ```
 
 A more advanced example that runs tests, go vet and Staticcheck on multiple OSs and Go versions looks like this:
--- a/website/content/docs/running-staticcheck/ci/github-actions/index.md
+++ b/website/content/docs/running-staticcheck/ci/github-actions/index.md
@@ -40,7 +40,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["windows-latest", "ubuntu-latest", "macOS-latest"]
-        go: ["1.17.x", "1.18.x"]
+        go: ["1.21.x", "1.22.x"]
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v4
--- a/website/content/docs/running-staticcheck/ci/github-actions/index.md
+++ b/website/content/docs/running-staticcheck/ci/github-actions/index.md
@@ -24,7 +24,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: dominikh/staticcheck-action@v1.3.0
         with:
-          version: "2023.1.6"
+          version: "2023.1.7"
 ```
 
 A more advanced example that runs tests, go vet and Staticcheck on multiple OSs and Go versions looks like this:
@@ -51,7 +51,7 @@ jobs:
       - run: "go vet ./..."
       - uses: dominikh/staticcheck-action@v1.3.0
         with:
-          version: "2023.1.6"
+          version: "2023.1.7"
           install-go: false
           cache-key: ${{ matrix.go }}
 ```
--- a/website/content/docs/running-staticcheck/ci/github-actions/index.md
+++ b/website/content/docs/running-staticcheck/ci/github-actions/index.md
@@ -18,15 +18,13 @@ on: ["push", "pull_request"]
 
 jobs:
   ci:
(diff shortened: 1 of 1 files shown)


novalagung:
friendly ping for @dominikh, kindly review this PR, thanks!

website/content/docs/running-staticcheck/ci/github-actions/index.md
```diff
@@ -18,15 +18,13 @@ on: ["push", "pull_request"]
...
-	- uses: dominikh/staticcheck-action@v1.2.0
-	  with:
-		version: "2022.1.1"
+    name: "Run CI"
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dominikh/staticcheck-action@v1.3.0
+        with:
+          version: "2022.1.1"
```
methane, line 27:
```suggestion
          version: "2023.1.6"
```

website/content/docs/running-staticcheck/ci/github-actions/index.md
```diff
@@ -37,27 +35,25 @@ on: ["push", "pull_request"]
...
+    steps:
+      - uses: actions/checkout@v4
+      - uses: WillAbides/setup-go-faster@v1.13.0
+        with:
+          go-version: ${{ matrix.go }}
+      - run: "go test ./..."
+      - run: "go vet ./..."
+      - uses: dominikh/staticcheck-action@v1.3.0
+        with:
+          version: "2022.1.1"
```
methane, line 54:
```suggestion
          version: "2023.1.6"
```

website/content/docs/running-staticcheck/ci/github-actions/index.md
```diff
@@ -37,27 +35,25 @@ on: ["push", "pull_request"]
...
-	  with:
-		version: "2022.1.1"
-		install-go: false
-		cache-key: ${{ matrix.go }}
+    name: "Run CI"
+    strategy:
+      fail-fast: false
+      matrix:
+        os: ["windows-latest", "ubuntu-latest", "macOS-latest"]
+        go: ["1.17.x", "1.18.x"]
```
ccoVeille, line 43: As it's for documentation purpose, I would suggest `["stable", " oldstable"]`

And if you disagree, to use 1.21.x and 1.22.x

novalagung: @ccoVeille I think using explicit version (1.21.x, 1.22.x) is better. I have applied the changes just now

ccoVeille: I'm OK with that.

On most project, and for most people, having fixed version is safer.

Some people are not aware of these, and if maintainer is not very active, every new PR could fail in the CI if something breaks the build by using a new version

ccoVeille (approved):
LGTM 👍

One remark anyway. Maybe I missed it, but I think this page should mention it's important to set up and use dependabot/renovate/whatever to keep this action update.

First. It would be better, and safer for everyone.

Then even if documentation is obsolete (because newer version of actions are available), it will be fixed automatically.

website/content/docs/running-staticcheck/ci/github-actions/index.md
ccoVeille: Maybe a yaml comment here, such as

# please adapt this to your need and recent version when available

dominikh:
> but I think this page should mention it's important to set up and use dependabot/renovate/whatever to keep this action update.

I do not advocate the use of Dependabot or similarly noisy tools that desire to update dependencies for the sake of updating them.

ccoVeille:
> > but I think this page should mention it's important to set up and use dependabot/renovate/whatever to keep this action update.
> 
> I do not advocate the use of Dependabot or similarly noisy tools that desire to update dependencies for the sake of updating them.

That's right. It's definitely noisy. But somehow useful, otherwise it won't be everywhere. But that's a another topic 😅

Stepping out about the need to use such a not, my remark stays. The documentation could mention the example provided in the markdown should mention it may be out of date, and people should adapt to use latest version.

dominikh:
I will probably switch to using `latest`, `stable` etc in the examples, and document that preferably people pin to specific versions instead. I'd much rather people ignore the advice to pin and run newer versions when they don't intend to than that people ignore the advice to update regularly and run older versions.

Plus I'm getting tired of having to update examples.

ccoVeille:
That's a good approach.

PR #1517, opened 2024-04-08
SA5008: check for missing closing quote

nanzhong:
This PR attempts to address handling the missing closing quote case for the struct tag parser.

I could be naive here, but changing the current implementation of the scanner from `break`ing when it's unable to find the closing double quote to returning an error would achieve this. This does the change the behaviour to now ignore partially valid tags (i.e. `` `xml:"complete" json:"missing` `` would no longer result in the `xml:"complete"` being checked), instead it will highlight the missing closing quote violation.

I'm not familiar with the intended `testdata` organization, so I've proposed a test case for this under the `example.com/CheckStructTags`.

### DIFF
--- a/staticcheck/sa5008/structtag.go
+++ b/staticcheck/sa5008/structtag.go
@@ -3,10 +3,12 @@
 
 package sa5008
 
-import "strconv"
+import (
+	"errors"
+	"strconv"
+)
 
 func parseStructTag(tag string) (map[string][]string, error) {
-	// FIXME(dh): detect missing closing quote
 	out := map[string][]string{}
 
 	for tag != "" {
@@ -43,7 +45,7 @@ func parseStructTag(tag string) (map[string][]string, error) {
 			i++
 		}
 		if i >= len(tag) {
-			break
+			return nil, errors.New("missing closing quote")
 		}
 		qvalue := string(tag[:i+1])
 		tag = tag[i+1:]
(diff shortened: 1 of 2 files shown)


arp242:
There's a bunch of other cases with quotes that should probably trigger an error; with your patch:

```
% git diff
diff --git i/staticcheck/sa5008/testdata/src/example.com/CheckStructTags/CheckStructTags.go w/staticcheck/sa5008/testdata/src/example.com/CheckStructTags/CheckStructTags.go
index 37b76813..966a6464 100644
--- i/staticcheck/sa5008/testdata/src/example.com/CheckStructTags/CheckStructTags.go
+++ w/staticcheck/sa5008/testdata/src/example.com/CheckStructTags/CheckStructTags.go
@@ -23,6 +23,14 @@ type T1 struct {
        O int        `json:"some-name,omitzero,omitempty,nocase,inline,unknown,format:'something,with,commas'"`
        P int        `json:"`          //@ diag(`missing closing quote`)
        Q int        `json:"some-name` //@ diag(`missing closing quote`)
+       R int        `json:some-name"`
+
+       AA int `json:"hmm""`
+       BB int `json:"hmm"""`
+       CC int `json:""hmm"`
+       DD int `json:"""hmm"`
+       EE int `json:""hmm""`
+       FF int `json:"""hmm"""`
 }

 type T2 struct {

% go test -count=1 ./staticcheck/sa5008
ok      honnef.co/go/tools/staticcheck/sa5008   0.578s
```

So no error on a missing starting quote or double quotes.

nanzhong:
@arp242 you're totally right, however, the reason those are resulting in a pass is because the current implementation of [`parseStructTag`](https://github.com/dominikh/go-tools/blob/master/staticcheck/sa5008/structtag.go#L8) intentionally silently stops parsing malformed struct tags when it encounters a part that does not follow the typical `key:"value"` format; note that all your test cases either are prefixed by a `/json:".*"/` or do not start with a valid sequence.

I suspect the intent here is to be conservative since the language spec doesn't go beyond specifying that struct tags must be a string literal value. I'm not sure how strong @dominikh wants to preserve this behaviour. If not, then I can certainly update the struct tag parsing to fail more loudly, if so, then perhaps I could introduce a separate optional check that's more strict about the format of struct tags?

PR #1542, opened 2024-05-25
SA1006: flag more functions

### LINKED ISSUE #1528: SA1006: flag more functions
We're at least missing `fmt.Errorf`, `log.Fatalf`, and several methods on `log.Logger` and methods from `testing`.

arp242:
Flag all stdlib functions that have both a Printf() and Print() variant.

Found with:

    % rg -g '!internal/' -g '!cmd/' '^func ([()*a-zA-Z0-9_ ]+)?[A-Z]\w+\(\w+ string, \w+ \.\.\.any\)'

(will need to filter out some false positives)

Fixes #1528

### DIFF
--- a/staticcheck/sa1006/sa1006.go
+++ b/staticcheck/sa1006/sa1006.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"go/ast"
 	"go/types"
+	"strings"
 
 	"honnef.co/go/tools/analysis/code"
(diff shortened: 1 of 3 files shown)


dominikh (approved):

arp242:
Added a second commit to special-case fixing fmt.Errorf() → errors.New() – previously that would get "fixed" to fmt.Error(). Just adding a note since you already approved this PR.

PR #1543, opened 2024-05-25
SA9009: flag defer foo() where foo returns a closure

### LINKED ISSUE #466: staticcheck: flag `defer foo()` where foo returns a closure
Inspired by https://github.com/golang/lint/issues/451

Obviously concerns about false positives exist. This check will need a thorough test run on the Go corpus.

arp242:
Add a new checker to flag cases where a deferred function returns exactly one function, and that function isn't called. For example:

	func f() func() {
		// Do stuff.
		return func() {
			// Do stuff
		}
	}

	func main() {
		defer f() // Error: should be f()()
	}

As mentioned in #466, there's a change for false positives. I ran the check on Go stdlib, all my own code, and about 200 popular Go packages, and haven't found any false positives (and one match: https://github.com/grpc/grpc-go/pull/7270)

Fixes #466


arp242:
I ran the same script for #1545 with this check; found three more wrong uses of defer, but no false positives.

philjb:
@arp242 - thanks for putting this pr together!

PR #1544, opened 2024-05-25
lintcmd: don't print \' in -list-checks and -explain

arp242:
All the single quotes are written as `\'` in the descriptions. I assume that's for a good reason, but -list-checks would print:

    % staticcheck -list-checks | grep "'"
    S1003 Replace call to \'strings.Index\' with \'strings.Contains\'
    S1004 Replace call to \'bytes.Compare\' with \'bytes.Equal\'
    S1011 Use a single \'append\' to concatenate two slices
    S1012 Replace \'time.Now().Sub(x)\' with \'time.Since(x)\'
    ...

    % staticcheck -explain S1003
    Replace call to \'strings.Index\' with \'strings.Contains\'

So fix that:

    % staticcheck -list-checks | grep "'"
    S1003 Replace call to 'strings.Index' with 'strings.Contains'
    S1004 Replace call to 'bytes.Compare' with 'bytes.Equal'
    S1011 Use a single 'append' to concatenate two slices
    S1012 Replace 'time.Now().Sub(x)' with 'time.Since(x)'
    ...

    % staticcheck -explain S1003
    Replace call to 'strings.Index' with 'strings.Contains'

As an added bonus, make `-explain s1003` (lower case) work. Seemed too small of a change to make a new PR for.

### DIFF
--- a/analysis/lint/lint.go
+++ b/analysis/lint/lint.go
@@ -155,25 +155,25 @@ func (doc *Documentation) format(markdown bool, metadata bool) string {
 	if markdown {
 		fmt.Fprintf(b, "%s\n\n", doc.TitleMarkdown)
 		if doc.Text != "" {
-			fmt.Fprintf(b, "%s\n\n", doc.TextMarkdown)
+			fmt.Fprintf(b, "%s\n\n", strings.TrimSpace(doc.TextMarkdown))
 		}
 	} else {
 		fmt.Fprintf(b, "%s\n\n", doc.Title)
 		if doc.Text != "" {
-			fmt.Fprintf(b, "%s\n\n", doc.Text)
+			fmt.Fprintf(b, "%s\n\n", strings.TrimSpace(doc.Text))
 		}
 	}
 
 	if doc.Before != "" {
 		fmt.Fprintln(b, "Before:")
 		fmt.Fprintln(b, "")
-		for _, line := range strings.Split(doc.Before, "\n") {
+		for _, line := range strings.Split(strings.TrimSpace(doc.Before), "\n") {
 			fmt.Fprint(b, "    ", line, "\n")
 		}
 		fmt.Fprintln(b, "")
 		fmt.Fprintln(b, "After:")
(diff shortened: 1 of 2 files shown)


dominikh:
FWIW, `\'` is how we encode backticks (for Markdown) in raw string literals.

arp242:
> As an added bonus, make -explain s1003 (lower case) work. Seemed too small of a change to make a new PR for.

I backed out this change, as per #1546. As I mentioned there, I thought all CLI flags were case-insensitive but they're not.

I also added a commit to print fewer extraneous newlines in the Before/After blocks; see the commit message.

arp242:
Yeah, I also discovered that on Sunday when I wanted to build the website. Fixing the `website/cmd/generate_checks` program wasn't too much work, but Hugo gives some error on shortcodes, and gave up at that.

dominikh:
Yeah, I'll have to fix `website` soon. Hugo has been "fun" when it comes to backwards compatibility.

arp242:
> Hugo has been "fun" when it comes to backwards compatibility.

I actually tried older versions of Hugo too, the current release from when the website was first added, and a release from a few months before that, but all gave the same error as the latest version:

```
% hugo
Start building sites …
hugo v0.126.2+extended linux/amd64 BuildDate=unknown

Total in 13 ms
Error: error building site: process: readAndProcessContent:
"/home/martin/src/go-tools/website/content/contact.md:9:1":
failed to extract shortcode: template for shortcode "blocks/section" not found
```

That was with v0.88.1 and v0.80.0, according to my shell history.

FYI.

dominikh:
Make sure you've cloned the repository recursively; the website theme is pulled in via a git submodule.

arp242:
Ah right, that furthers things! Got fresh errors though.

I just wanted to look at some of the documentation issues in the weekend, and make sure whatever I wrote appears well on the website. Well, whatever 🤷

The amount of complexity involved here for rendering a few basic HTML pages is slightly ridiculous to be honest.

PR #1545, opened 2024-05-25
SA1032: add check for errors.Is() and .As() argument order

### LINKED ISSUE #1530: New check: Warn when errors.Is() arguments swapped
It's almost never right to write `errors.Is(fs.ErrNotExist, err)`, but I still do it sometimes to much head-scratching :sweat_smile: It should be `errors.Is(err, fs.ErrNotExist)`.

Would love it if there's way to warn when the first argument is an exported error value, especially one in another package... catch it much sooner.

Thanks for the consideration!

Also, many thanks for Staticcheck. It is a huge timesaver.

arp242:
Fixes #1530


arp242:
I ran this against about 400 modules; no errors.

Full list: https://gist.github.com/arp242/6251e33e9912882da67b3e025cf948c4

Issue #1526, opened 2024-04-23
Detect defer calls skipped by os.Exit


fionera:
The following code is not executing its defer function as log.Fatal calls syscall.Exit. This can prevent cleanup functions to run and could be unexpected to some developers.

```go
package main

import (
    "log"
)

func main() {
    defer func() {
        println("no worries I will help you, too bad this doesn't execute")
    }()

    log.Fatal("oh no I slipped")
}
```

arp242:
This has definitely taken me by surprise on a few occasions; the question is: how often is it not a problem that the defer doesn't run? I fear such a check might be very noisy.

Below is a quick (incomplete) patch; I ran it on my set of test code and it didn't flag anything, but that's mostly libraries rather than applications, which shouldn't call os.Exit() or log.Fatal() in the first place. Also not 100% sure my patch below will actually catch all cases.

Someone should probably run this on a whole bunch of `package main` applications and see what comes rolling out.

```
diff --git c/staticcheck/analysis.go w/staticcheck/analysis.go
index cce61975..b682a3aa 100644
--- c/staticcheck/analysis.go
+++ w/staticcheck/analysis.go
@@ -97,6 +97,7 @@ import (
 	"honnef.co/go/tools/staticcheck/sa9006"
 	"honnef.co/go/tools/staticcheck/sa9007"
 	"honnef.co/go/tools/staticcheck/sa9008"
+	"honnef.co/go/tools/staticcheck/sa9010"
 )
 
 var Analyzers = []*lint.Analyzer{
@@ -193,4 +194,5 @@ var Analyzers = []*lint.Analyzer{
 	sa9006.SCAnalyzer,
 	sa9007.SCAnalyzer,
 	sa9008.SCAnalyzer,
+	sa9010.SCAnalyzer,
 }
diff --git c/staticcheck/sa9010/sa9010.go w/staticcheck/sa9010/sa9010.go
new file mode 100644
index 00000000..e5b381b3
--- /dev/null
+++ w/staticcheck/sa9010/sa9010.go
@@ -0,0 +1,93 @@
+package sa9010
+
+import (
+	"fmt"
+	"go/types"
+
+	"golang.org/x/tools/go/analysis"
+	"honnef.co/go/tools/analysis/lint"
+	"honnef.co/go/tools/analysis/report"
+	"honnef.co/go/tools/go/ir"
+	"honnef.co/go/tools/go/ir/irutil"
[...]

PR #1546, opened 2024-05-26
lintcmd: make -check case-insensitive, error on unknown checks

arp242:
This changes it so that the check is case-insensitive, and so it will return an error on unknown checks. Previously:

	% staticcheck -checks=sa1000 ./lintcmd
	[no output, exit 0]

	% staticcheck -checks=banana ./lintcmd
	[no output, exit 0]

	% staticcheck -checks=SA1000 ./lintcmd
	lintcmd/cmd.go:270:34: error parsing regexp: missing closing ): `$^(` (SA1000)

This is massively confusing! When I was doing my testing for the other PRs yesterday I found it odd that I didn't get any error for the defer one, since I *knew* there was at least one module with this error. Turned out it just always passed because I used lower-case.

I wouldn't be surprised if tons of people have made this mistake and have run (or are running) staticcheck runs that don't actually do anything.

All of this also applies to checks = [..] from the config file.

### DIFF
--- a/lintcmd/cmd.go
+++ b/lintcmd/cmd.go
@@ -678,7 +678,11 @@ func (cmd *Command) printDiagnostics(cs []*lint.Analyzer, diagnostics []diagnost
 	for i, a := range cs {
 		analyzerNames[i] = a.Analyzer.Name
 	}
-	shouldExit := filterAnalyzerNames(analyzerNames, fail)
+	shouldExit, err := filterAnalyzerNames(analyzerNames, fail)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "%s\n", err)
+		return 1
+	}
 	shouldExit["staticcheck"] = true
 	shouldExit["compile"] = true
 
--- a/lintcmd/lint.go
+++ b/lintcmd/lint.go
@@ -10,6 +10,7 @@ import (
 	"os/signal"
 	"path/filepath"
 	"regexp"
+	"slices"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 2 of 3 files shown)


dominikh:
> and so it will return an error on unknown checks

We intentionally don't complain on unknown checks, at least in the `staticcheck.conf` case. This is both for backwards and forwards compatibility, as well as to accommodate projects that use custom builds of Staticcheck that include custom checks. In neither case do we want to prevent people with other versions of Staticcheck from running Staticcheck. And I'd like to keep `-checks` and the config behaving the same, to avoid more confusion.

Maybe as a compromise we can emit diagnostics for unknown checks?

> This changes it so that the check is case-insensitive

I can see the appeal in that, but for consistency we'd also have to do this for `//lint:ignore`. Then we'd probably also have to update gopls to interpret its [`analyses`](https://github.com/golang/tools/blob/master/gopls/doc/settings.md#-analyses-mapstringbool) setting, and after all that we'd still be left with case-sensitive anchors on [our website](https://staticcheck.dev/docs/checks/), as well as other analysis runners that aren't case-insensitive.

I think I'd prefer it if we stuck to only one way of spelling check names (i.e. the current way), and instead focused on making it more obvious when check names are misspelled.

arp242:
Emitting a diagnostic is okay. I'll change that.

Personally I'd strongly prefer case-insensitive checks whenever feasible. To me at least it's much more natural to type "sa1000" rather than "SA1000" (even when *knowing* about this, I again [did it wrong yesterday](https://github.com/dominikh/go-tools/issues/1526#issuecomment-2132413263)).

It should probably be case-insensitive in `//lint:` directives too; I don't mind working on changing that.

We can't make everything 100% case-insensitive, and that's okay; not many people are manually typing anchor tags for the website. If we can do >99% of where people are typing it then that's a win IMO. We can know what's intended with 100% certainty, so it just doesn't make any sense to stand on ceremony and reject it.

This is also why I made `-explain` case-insensitive in #1544, because I kept typing `-explain sa1005`, getting and error, and having to re-do it (when I made that PR I thought `-checks` was case-insensitive, as it didn't error).

PR #1548, opened 2024-05-26
s1009: fix with SelectorExpr

### LINKED ISSUE #1527: No check for unnecessary `nil` check on a map followed by a `len` check of the same map
Considering staticcheck identifies unnecessary nil checks for slices followed by a `len` check, I am surprised it doesn’t seem to do the same for maps.

I was just reviewing some code that is essentially:

```
type obj struct {
    m map[uint64]bool
}

func (o *obj) method() {
    if o.m != nil && len(o.m) > 0 {
        // start goroutine to process elements of `o.m`
    }
}
```

I was expecting staticcheck to flag the o.m != nil part as unnecessary, but it isn’t.

arp242:
Also clarify the docs that this also checks maps and channels.

Fixes #1527


arp242:
Ran this against my test set of ~700 modules; flagging a bunch of cases correctly. Don't see anything odd.

PR #1549, opened 2024-05-27
ST1020: allow comments to start with deprecation comment

### LINKED ISSUE #1378: ST1020: ignore deprecation comment
As `ST1020` is opt-in I would expect it to be ignored if the only comment on an exported function is a deprecation message:
```
% txtar-c .
-- foo.go --
package foo

// Deprecated: Use bar instead
func Foo() {
}
-- go.mod --
module x

go 1.21
% staticcheck -checks=ST1020 .
foo.go:3:1: comment on exported function Foo should be of the form "Foo ..." (ST1020)

### PR DESCRIPTION
**Pull request by arp242**
Fixes #1378

### DIFF
--- a/stylecheck/st1020/st1020.go
+++ b/stylecheck/st1020/st1020.go
@@ -55,6 +55,10 @@ func run(pass *analysis.Pass) (interface{}, error) {
 		if !ast.IsExported(decl.Name.Name) {
 			return
 		}
+		if strings.HasPrefix(text, "Deprecated: ") {
+			return
+		}
+
 		kind := "function"
 		if decl.Recv != nil {
 			kind = "method"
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: stylecheck/st1020/st1020.go [hunk 1]
```diff
@@ -55,6 +58,10 @@ func run(pass *analysis.Pass) (interface{}, error) {
 		if !ast.IsExported(decl.Name.Name) {
 			return
 		}
+		if _, ok := depr.Objects[pass.TypesInfo.ObjectOf(decl.Name)]; ok && strings.HasPrefix(text, "Deprecated:") {
```
dominikh, line 61: When would a comment have the right prefix but not be marked as deprecated by `deprecated.Analyzer`?

arp242: I put that strings.HasPrefix in so that this:

    // wrong comment yo. //@diag (`comment on exported function`)
    //
    // Deprecated: don't use.
    func F6() {}

Still gets flagged, and it only skips if the `Deprecated: ...` comment is the *only* comment.

dominikh: I'm not confused about the `strings.HasPrefix`, but about the use of `depr`. Isn't the `strings.HasPrefix` call enough?

arp242: Oh right. Yeah, I guess that it rather redundant an pointless; just seemed like the "proper way" to do it. I removed it.

PR #1555, opened 2024-05-31
SA4006: Fix for compound operators, increment/decrement

### LINKED ISSUE #1329: SA4006: detect unused writes with +=
```
$ staticcheck -debug.version
staticcheck 2022.1.3 (v0.3.3)

Compiled with Go version: go1.18.4
Main module:
        honnef.co/go/tools@v0.3.3 (sum: h1:oDx7VAwstgpYpb3wv0oxiZlxY+foCpRAwY7Vk6XpAgA=)
Dependencies:
        github.com/BurntSushi/toml@v0.4.1 (sum: h1:GaI7EiDXDRfa8VshkTj7Fym7ha+y8/XxIgD2okUIjLw=)
        golang.org/x/exp/typeparams@v0.0.0-20220218215828-6cf2b201936e (sum: h1:qyrTQ++p1afMkO4DPEeLGq/3oTsdlvdH4vqZUBWzUKM=)
        golang.org/x/mod@v0.6.0-dev.0.20220419223038-86c51ed26bb4 (sum: h1:6zppjxzCulZykYSLyVDYbneBfbaBIQPYMevg0bEwv2s=)
        golang.org/x/sys@v0.0.0-20211019181941-9d821ace8654 (sum: h1:id054HUawV2/6IGm2IV8KZQjqtwAOo2CYlOToYqa0d0=)
        golang.org/x/tools@v0.1.11-0.20220513221640-090b14e8501f (sum: h1:OKYpQQVE3DKSc3r3zHVzq46vq5YH7x8xpR3/k9ixmUg=)
```

Go version is 1.19.2 linux/amd64.

Example program:

```
package main

func main() {
	dir, err := f()
	if err != nil {
		return
	}
	dir += "/a"
}

func f() (string, error) { return "", nil }
```

Staticcheck doesn't complain about that. But if I tweak it to

```
package main

func main() {
	dir, err := f()
	if err != nil {
		return
	}
	dir = dir + "/a"
}

func f() (string, error) { return "", nil }
```

Then I get

    x/x.go:8:2: this value of dir is never used (SA4006)

Could SA4006 better detect this case with `+=`?

arp242:
In the case of compound operators such as +=, -=, etc we need to use the left-hand side to get the value.

For IncDec we need a different codepath from AssignStmt altogether.

Fixes #1329


cespare:
Thanks for implementing this @arp242! I skimmed the code and it makes sense to me, but I'm not too familiar with staticcheck internals since all my tools are built on top of the x/tools code.

I ran this over a large code base and it worked fine and flagged one new unused write (thanks!).

I tried running it against another instance of this bug that I found in a code review recently and it wasn't flagged by this code. But after some digging I see that it's a general shortcoming of the SA4006 implementation, not this PR; I filed #1609.

Hopefully this can be incorporated into staticcheck soon.

PR #1580, opened 2024-07-31
QF1013: detect strings that could be rewritten with a string literal

### LINKED ISSUE #1566: quickfix: detect quoted string escaping quote, while backticks could be used
I didn't find a linter for this so I'm creating an issue here.

I asked Gopher Slack community https://gophers.slack.com/archives/C4MTR76PQ/p1719215628975379

And it appears I was right and this linter/checker doesn't exist yet.

The idea is to be able to report this for any string

```go
fmt.Println("I'm \"happy\"")
```

```go
fmt.Println(`I'm "happy"`)
```

This should be an optional rule of course, as it could bring a lot of noise.

I think one could improve the readability of the code

ccoVeille:
fix #1566


ccoVeille:
Here is a PR that implements the conversion of string to string literal as discussed in 
- https://github.com/dominikh/go-tools/issues/1566

The code adapt what is done in https://github.com/dominikh/go-tools/blob/master/simple/s1007/s1007.go and https://github.com/dominikh/go-tools/blob/master/stylecheck%2Fst1018%2Fst1018.go

ccoVeille:
The current code is working.

I'm asking for a code review, but also some advice about this code. because it will somehow conflict with what is done in 

 https://github.com/dominikh/go-tools/blob/master/simple/s1007/s1007.go 

as s1007 and q1013 may report twice the error

what would be the best thing to do ?

Thanks

dominikh:
> as s1007 and q1013 may report twice the error

That's probably not much of a problem, especially because QF checks are only used in gopls, not the `staticcheck` CLI tool, so at worst we'll get two suggestions that fix the same problem. If gopls cares, they can filter the results.

I'll get to review this PR eventually, but it might take a while, as my backlog is fairly long at the moment.

ccoVeille:
Noted, thanks

ccoVeille:
@antonboom

I saw your comment [here](https://github.com/dominikh/go-tools/issues/1566#issuecomment-2308884698)

Maybe you could review quickly.

I'm curious about your point of view about the fact it should be only a quickfix (not not reported via staticcheck, so not reported by golangci-lint)

ccoVeille:
@dominikh do you think you could take a look at my old PR?

I think gopls is the right candidate, but if things are moving I might open a PR to add this to revive linters

dominikh:
Sorry, but I don't have the capacity for that. Please see https://github.com/dominikh/.github/blob/main/CONTRIBUTING.md#pull-requests for some background.

PR #1585, opened 2024-08-13
analysis/lint/testutil: unset GO111MODULE in test

zhsj:
In Debian packaging system, GO111MODULE is set to off. This causes various test failures due to dialog message mismatch.

### DIFF
--- a/analysis/lint/testutil/util.go
+++ b/analysis/lint/testutil/util.go
@@ -93,7 +93,7 @@ func Run(t *testing.T, a *lint.Analyzer) {
 			cfg := &packages.Config{
 				Dir:   dir,
 				Tests: true,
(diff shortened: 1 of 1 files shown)


dominikh:
I am not sure why the tests are passing in CI, but we shouldn't be able to set GO111MODULE=off, as the test suite relies on Go modules to test different language versions.

zhsj:
> I am not sure why the tests are passing in CI, but we shouldn't be able to set GO111MODULE=off, as the test suite relies on Go modules to test different language versions.

The tests didn't pass with GO111MODULE=off. So this PR ensures such unwanted env is unset.

dominikh:
Oh, I see. I misread your change. Thanks!

PR #1620, opened 2024-11-28
Refactored printf package

Turtel216:
I found some readability issues in the ``printf`` package, mainly nested code, one letter variables and poor organization. I refactored the package to address these issues. Furthermore, I made sure that all tests pass without any errors


dominikh:
Thanks, but this is unnecessary churn and not an improvement.

PR #1643, opened 2025-04-02
Update sa6005.go, give a better tips

alingse:
```go
xyz.go:55:13: SA6005: should use strings.EqualFold instead (staticcheck)
} else if strings.ToUpper(a) != strings.ToUpper(b) {
```

this message may cause some misstake when copy the method and change it to

```go
} else if strings.EqualFold(a, b) {
```


alingse:
@dominikh  how do you think? 🤔

PR #1663, opened 2025-07-20
staticcheck: add documentation text to select checks

catzkorn:
This PR introduces documentation text for the following checks:

- SA1002
- SA1012
- SA1014
- SA1020
- SA2000
- SA2003
- SA4010

It also updates the format of the method reference in the `Title` string to be consistent with other parts of the staticcheck
documentation and the Go community conventions.

### DIFF
--- a/staticcheck/sa1002/sa1002.go
+++ b/staticcheck/sa1002/sa1002.go
@@ -21,6 +21,7 @@ var SCAnalyzer = lint.InitializeAnalyzer(&lint.Analyzer{
 	},
 	Doc: &lint.RawDocumentation{
 		Title:    `Invalid format in \'time.Parse\'`,
+		Text:     `\'time.Parse\' requires a layout string that uses Go's reference time: \'Mon Jan 2 15:04:05 MST 2006\' (Unix date format). The layout must represent this date and time exactly. See https://pkg.go.dev/time#pkg-constants for layout examples.`,
 		Since:    "2017.1",
 		Severity: lint.SeverityError,
 		MergeIf:  lint.MergeIfAny,
--- a/staticcheck/sa2003/sa2003.go
+++ b/staticcheck/sa2003/sa2003.go
@@ -21,6 +21,7 @@ var SCAnalyzer = lint.InitializeAnalyzer(&lint.Analyzer{
 	},
 	Doc: &lint.RawDocumentation{
 		Title:    `Deferred \'Lock\' right after locking, likely meant to defer \'Unlock\' instead`,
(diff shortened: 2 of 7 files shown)


staticcheck/sa4010/sa4010.go
```diff
@@ -17,6 +17,7 @@ var SCAnalyzer = lint.InitializeAnalyzer(&lint.Analyzer{
 	},
 	Doc: &lint.RawDocumentation{
 		Title:    `The result of \'append\' will never be observed anywhere`,
+		Text:     `Calls to \'append\' produce a new slice value that must be used. When the result of \'append\' is assigned to a variable that is never subsequently read, or is immediately overwritten, the append operation has no effect.`,
```
dominikh, line 20: > immediately overwritten

doesn't have to be immediate, just before it's read.

> append operation has no effect

It may have an effect, but it usually won't be the desired one.

staticcheck/sa1002/sa1002.go
```diff
@@ -21,6 +21,7 @@ var SCAnalyzer = lint.InitializeAnalyzer(&lint.Analyzer{
 	},
 	Doc: &lint.RawDocumentation{
 		Title:    `Invalid format in \'time.Parse\'`,
+		Text:     `\'time.Parse\' requires a layout string that uses Go's reference time: \'Mon Jan 2 15:04:05 MST 2006\' (Unix date format). The layout must represent this date and time exactly. See https://pkg.go.dev/time#pkg-constants for layout examples.`,
```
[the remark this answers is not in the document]
dominikh: Use `\"` instead of `\'` to quote the reference time so we get quotes in CLI output.

staticcheck/sa2003/sa2003.go
```diff
@@ -21,6 +21,7 @@ var SCAnalyzer = lint.InitializeAnalyzer(&lint.Analyzer{
 	},
 	Doc: &lint.RawDocumentation{
 		Title:    `Deferred \'Lock\' right after locking, likely meant to defer \'Unlock\' instead`,
+		Text:     `Deferring \'Lock\' immediately after locking will cause a deadlock. Use \'Unlock\' instead.`,
```
dominikh, line 24: It doesn't necessarily cause a deadlock (and there are indeed possible false positives with this check if you write hard to follow code). 

At the end of the day, this check just flags a probable typo, I'm not sure what we could document about that.

dominikh (changes requested):

dominikh:
Thanks! I went ahead and slightly edited the documentation of SA1002.

PR #1667, opened 2025-08-02
Add ST1024: Avoid named error return values named 'err'

thaJeztah:
Named return values called 'err' are discouraged because they can lead to confusion and subtle mistakes.

In most Go code, 'err' is used for short-lived local variables. Naming a return parameter 'err' can induce cognitive load.

For example:

	func fn() (err error) {
		if err := doSomething(); err != nil { // shadows the return
			return err
		}
		defer func() {qqq
			// This 'err' is the named return, not the one above
			if err != nil {
				log.Println("deferred error:", err)
			}
		}()
		return nil
	}

Using a distinct name for the named return makes the return value’s role explicit and avoids confusion:

	func fn() (retErr error) {
		if err := doSomething(); err != nil {
			return err
		}
		defer func() {
			if retErr != nil {
				log.Println("deferred error:", retErr)
			}
		}()
		return nil
	}

This check is opt-in and not enabled by default.


thaJeztah:
@dominikh ptal; I'm really unfamiliar with this codebase, so hope I did it correct (tests passed locally, so that's a win 😅).  This was a bit of a pet-peeve of mine (`err` for named outputs), and I didn't know of any linter that had a check for this, so thought I'd give it a try, taking parts of https://github.com/dominikh/go-tools/tree/master/stylecheck/st1008 as an example.

I hope it's something you're willing to accept (I think I made it "opt-in"), but happy to hear your thought! (if not accepted, it was a fun exercise 😂)

PR #1677, opened 2025-09-14
SA1027: add s390x support to test file

QuLogic:
This fails to find any test data without the `s390x` build tag, and the test would fail.


dominikh:
Thanks

PR #1690, opened 2025-12-10
pattern: run lexer in a lightweight coroutine

adonovan:
Previously, the lexer ran as a separate goroutine that fed a channel with items. This change expresses the lexer as an iter.Seq[item], and then uses the iter.Pull mechanism to run it as a lightweight coroutine. This reduces the running time of the added benchmark by about 32%.

$ go test  -benchtime=10s -bench=. ./pattern

 before
 BenchmarkParser-8   	  649060	     19051 ns/op
 after
 BenchmarkParser-8   	  908734	     12971 ns/op

Interestingly, the total time spent by gopls in Parse, measured by bracketing Parse with time.Now/time.Time.Sub calls, consistently reduces by a factor of around 20x (down to 800us, originally 15ms). I can't yet explain this discrepancy.

### DIFF
--- a/pattern/parser.go
+++ b/pattern/parser.go
@@ -363,10 +363,6 @@ func (p *Parser) bindingIndex(name string) int {
 }
 
 func (p *Parser) Parse(s string) (Pattern, error) {
-	t0 := time.Now()
-	defer func() {
-		log.Printf("pattern.Parse(%d, %32q) = %v", len(s), s, time.Since(t0))
-	}()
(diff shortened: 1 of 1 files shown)


dominikh:
Those gopls timings are quite worrying...

As for the change itself, maybe we should just have the lexer return a slice of tokens? Patterns are small, we only parse them at program startup, and the [cg]oroutines probably don't save enough garbage to justify the added complexity.

Issue #1684, opened 2025-10-10
QF1012 : b.WriteString(x + y) -> b.WriteString(s); b.WriteString(y)


adonovan:
The checker that replaces b.WriteString(fmt.Sprintf(...)) with Fprintf(&buf, ...) could also optimize calls to WriteString where the operand is a string concatenation. Replacing it with two calls to WriteString  would avoid an allocation.

IlyasYOY:
@dominikh I've opened a PR with the proposed change. I marked it as Draft so you can think: 

- if the change fits the existing parser and should not be extracted separately,
- maybe I've already did some no-goes.

If you think it look ok then I'll continue fixing open conversations.

Thank you!

PR #1697, opened 2026-02-06
QF1012 : b.WriteString(x + y) -> b.WriteString(s); b.WriteString(y) #1684

IlyasYOY:
Adds support for a simple string concat into QF1012 as mentioned in https://github.com/dominikh/go-tools/issues/1684.

I'd be glad to provide the fixes you think are necessary.

### DIFF
--- a/quickfix/qf1012/qf1012.go
+++ b/quickfix/qf1012/qf1012.go
@@ -54,11 +54,19 @@ var (
 				(Symbol "fmt.Sprintf")
 				(Symbol "fmt.Sprintln"))
 			args))`)
+
+	checkWriteStringConcatQ = pattern.MustParse(`
+	(CallExpr
+		(SelectorExpr recv (Ident "WriteString"))
+		(BinaryExpr
+			left@(BasicLit "STRING" _)
+			"+"
+			right@(Ident _)))`)
 )
 
 func run(pass *analysis.Pass) (any, error) {
 	fn := func(node ast.Node) {
-		getRecv := func(m *pattern.Matcher) (ast.Expr, types.Type) {
+		getRecv := func(m *pattern.Matcher) (ast.Expr, ast.Expr, types.Type) {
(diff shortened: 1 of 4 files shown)


quickfix/qf1012/qf1012.go
```diff
@@ -115,14 +120,65 @@ func run(pass *analysis.Pass) (any, error) {
 					X:   ast.NewIdent("fmt"),
 					Sel: ast.NewIdent(newName),
 				},
-				Args: append([]ast.Expr{recv}, args...),
+				Args: append([]ast.Expr{recvPtr}, args...),
 			}))
 			report.Report(pass, node, msg, report.Fixes(fix))
+		} else if m, ok := code.Match(pass, checkWriteStringConcatQ, node); ok {
+			_, recv, recvT := getRecv(m)
+			// TODO(IlyasYOY): add support for []bytes, similar to the block above.
```
IlyasYOY: I will definitely need to add `Write` method handling here.

quickfix/qf1012/qf1012.go
```diff
@@ -115,14 +120,65 @@ func run(pass *analysis.Pass) (any, error) {
...
+		ys := unwrapRight(bin.Y)
+
+		all := make([]*ast.BasicLit, 0, len(xs)+len(ys))
+		all = append(all, xs...)
+		all = append(all, ys...)
+
+		return all
+	}
+
+	// TODO(IlyasYOY): handle case for any expression returning string.
```
IlyasYOY: I will refactor this logic to handle more cases. For now, I support only literal ones.

Maybe you have an idea which cases I might beware the most?

IlyasYOY (commented):
I'd be glad to finish this PR, if you think I'm heading in the right direction and you want to see this feature here.

dominikh:
I'll try to take a look at this once I get the chance, but it might be a while.

PR #1703, opened 2026-03-03
s1034: Add explanatory example

indrora:
There is currently no explanation for this particular check. This change adds a simple example of this pattern.


indrora:
Bump for visibility.

dominikh:
I'm sorry, but as is, the explanation is wrong. To work correctly, the second type switch needs a short variable declaration, see https://tip.golang.org/ref/spec#Type_switches, but then the explanation of types being "inferred" would still not align with Go's semantics.

As per the [contributing guidelines](https://github.com/dominikh/.github/blob/main/CONTRIBUTING.md) I'd prefer to eventually write the documentation myself.

Thank you and I'm sorry.

PR #1704, opened 2026-03-06
lintcmd/cache: implement GOCACHEPROG

### LINKED ISSUE #1458: GOCACHEPROG or otherwise hookable caching support
Go 1.21 added `GOEXPERIMENT=cacheprog` to let you set GOCACHEPROG to the name of a tool to run as a child process and implement the cmd/go action/build cache via a JSON API over stdin/stdout. (https://github.com/golang/go/issues/59719)

We (@tailscale) are moving to using this in our CI to share a build cache over multiple machines without the slow tar/untar thing that people typically do on GitHub Actions. So far it's working great.

Except staticcheck has its own on-disk cache (a fork of cmd/go's cache prior to GOCACHEPROG support), so we can't share a cache for staticcheck between runs without backing up & restoring the cache directory, or using NFS/SMB/9P/FUSE to a shared cache server.

@dominikh, would you be amenable to us either adding GOCACHEPROG support to staticcheck (admittedly perhaps premature, considering it's still in GOEXPERMENT status), or at least a hook in the staticcheck code to let us build our own staticcheck binary that implements the hook via HTTP client calls to our cache server?

/cc @maisem

bradfitz:
Fixes dominikh/go-tools#1458

### DIFF
--- a/analysis/lint/testutil/util.go
+++ b/analysis/lint/testutil/util.go
@@ -67,7 +67,7 @@ func Run(t *testing.T, a *lint.Analyzer) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	c.SetSalt(salt)
+	cache.SetSalt(salt)
 
 	tags := build.Default.ReleaseTags
 	maxVersion := tags[len(tags)-1]
--- a/go/loader/hash.go
+++ b/go/loader/hash.go
@@ -13,8 +13,8 @@ import (
 // computeHash computes a package's hash. The hash is based on all Go
 // files that make up the package, as well as the hashes of imported
 // packages.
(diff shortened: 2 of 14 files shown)


dominikh:
Applied as ba2c20843b3252c9a2df60c2d740fc56d06208da. As for integration tests, that'll be my users (you) for now.

bradfitz:
Thanks! Would you like integration tests? If so, how much are you okay with dependency-wise? Would you be okay depending on https://github.com/bradfitz/go-tool-cache for a test server, or prefer a copy?

dominikh:
I wouldn't mind some tests, and a dependency would be fine :+1:

Issue #1712, opened 2026-04-20
SA1026: false positive for fields in embedded struct that are overridden


arp242:
The following code:

```
func main() {
    resp := &http.Response{
        StatusCode: 200,
        // ... etc
    }

    j, err := json.Marshal(struct {
        *http.Response
        Body    []byte
        Request *http.Request
    }{Response: resp, Body: []byte("ABC")})

    if err != nil {
        panic(err)
    }

    fmt.Println(string(j))
}
```

Works as expected; the `http.Request` struct contains some fields that can't be marshaled (func, channel), but it overrides the Request field:

```
% go run test.go
{
   [..]
   "StatusCode": 200,
   [..]
}
```

But staticcheck errors:

    % staticcheck test.go
    test.go:15:25: trying to marshal unsupported type func() (io.ReadCloser, error), via x.Request.GetBody (SA1026)

staticcheck (devel, v0.7.0-0.dev.0.20260409232628-ba2c20843b32)

dominikh:
Not embedding related, this is a simpler reproducer:

```
func main() {
	resp := &http.Response{
		StatusCode: 200,
		// ... etc
	}

	j, err := json.Marshal(resp)
	if err != nil {
		panic(err)
	}

	fmt.Println(string(j))
}
```

The problem is that our analysis is type-based, but you can marshal a `*http.Request` as long as it is nil, because the encoder won't recurse into it to find the field of unsupported type.

I don't think we can avoid this false positive in the general case (with non-trivial data flow) without neutering the check a fair bit.

PR #1713, opened 2026-04-20
SA1026, SA9005: check MarshalIndent functions

### LINKED ISSUE #1264: SA1026, SA9005: check MarshalIndent functions
These checks verify json.Marshal, xml.Marshal, and related functions, but they are missing at least json.MarshalIndent.

arp242:
While extracting a small test case from my real code for #1712, I suddenly could not longer reproduce it. Turns out that was because I used json.MarshalIndent() for my test case, and this function wasn't checked.

Fixes #1264


dominikh:
Can you drop the commit bumping actions/checkout? The CI failure was transient.

PR #1722, opened 2026-06-23
SA5008: add support for `embed` tag option

dsnet:
In golang/go#79985, it was decided to rename `inline` as `embed` to avoid compatibility issues with existing appearances of the `inline` tag option.

Add `embed` as a tag option understood by the tool.


dsnet:
\cc @dominikh

dominikh:
Thanks, will merge soon

PR #1726, opened 2026-07-13
staticcheck: fix nilness panic on recover

blampe:
Fixes #1725.

(I hope you're doing well, Dominik!)

PR #1728, opened 2026-07-23
nilness: bound generated state machine analysis

zbarsky-openai:
Dense nilness dataflow can spend more than 263 seconds analyzing one generated go-json encoder, even after the source carries the standard generated-file marker. The observed case was the vm_color_indent nogo action in a focused downstream build; the full invocation was interrupted after 286 seconds.

For functions with more than 1,000 IR blocks, check that the function belongs to a generated file before using the existing conservative unknown-return bailout. All handwritten functions and smaller generated functions retain their current nilness analysis. The new focused regression test constructs the large generated state machine at runtime instead of committing a large fixture.

This is intentionally separate from the recover fix in #1726.


dominikh:
Thank you, but the number of blocks is a poor proxy for CFG complexity. There is also a better way to fix this, by improving the data flow analysis runner. `github.com/goccy/go-json/internal/encoder/vm_color_indent` is only slow to analyze because it is pathological for the current runner. I have a local change that fixes this that I'll push once it is ready.

PR #1730, opened 2026-08-06
s1024: don't suggest an invalid fix when Sub is not called on a time.Time

### LINKED ISSUE #756: S1024: incorrect suggested fix for embedded type
This reproduces with Staticcheck 2020.1.4 (built into `gopls`), `go version` is `go1.14.1 darwin/amd64`.

Repro:

```go
package main

import "time"

type embedTime struct {
	time.Time
}

func main() {
	t := embedTime{}
	t.Sub(time.Now())
}
```

The suggested fix will convert `t.Sub(time.Now())` to `time.Until(t)`, which does not compile. The correct fix would be `time.Until(t.Time)`.

Eljees:
Fixes #756.

S1024 rewrites `x.Sub(time.Now())` into `time.Until(x)`, which only compiles when `x` is itself a `time.Time`. `Sub` is reachable in two other shapes, and the fix produces code that does not build in either:

```go
type embedTime struct{ time.Time }

func fnEmbedded(t embedTime) { t.Sub(time.Now()) } // fix: time.Until(t)
func fnPointer(t *time.Time) { t.Sub(time.Now()) } // fix: time.Until(t)
```

The pointer case is not in the issue, but it has the same cause.

Emitting `time.Until(t.Time)` for the embedded case would be nicer than dropping the fix, but the embedded field is not always nameable at the use site — it can be unexported and declared in another package — so this keeps the diagnostic and reports it without a fix instead.

The testdata gains both shapes. On the current tree the golden file fails with exactly the two invalid rewrites.


Eljees:
Checking in — this has been open since 6 August with no review yet. It's green and mergeable; happy to rebase or add anything you'd like before you take a look.

dominikh:
Sorry, but I'm slow to review PRs. Nothing is needed from your side at this time.

PR #1731, opened 2026-08-06
st1018: don't suggest escape sequences inside raw string literals

### LINKED ISSUE #1115: ST1018: fix and improve handling of raw string literals
Currently, Staticcheck emits a suggested fix that is invalid for raw string literals, as `\u` escape sequences don't work inside them.
While we fix that, we should adjust the diagnostic message, too. We shouldn't tell the user to use escape sequences in raw strings.

Eljees:
Fixes #1115.

Escape sequences are not interpreted inside raw string literals, so replacing an invisible character there with its escape sequence changes the string's value to the six literal characters of that sequence rather than removing the invisible one. The current testdata golden file records exactly that: the `e = ...` line holds a raw literal whose zero-width space has been "replaced" by six visible characters.

The message has the same problem: it tells the user to use an escape sequence that would not work where they are.

For raw string literals this now offers only the deletion fix, and the message points at an interpreted string literal instead. On a file with the same zero-width space in both kinds of literal:

```
main.go:6:17: string literal contains the Unicode format character U+200B, consider using the ... escape sequence instead
main.go:7:9:  raw string literal contains the Unicode format character U+200B, consider using an interpreted string literal and the ... escape sequence instead
```

The golden file loses its `replace format character U+200B with ...` section, which the test flags on its own: without that removal, `TestTestdata` fails with "golden has section for suggested fix ..., but we didn't produce any fix by that name".

I did not add testdata rows for a raw literal holding several invisible characters, since every added line rewrites all eight golden sections; happy to do it if you'd rather have the coverage.

(Edited: the first version of this description contained literal zero-width characters in the places where the escape sequence was meant to be shown, which rather made the point.)


Eljees:
Marking this ready for review. I'd been holding it as a draft while #1730 was open, in case you'd rather take contributions one at a time — say the word and I'll close one. A draft doesn't show up in your review queue at all, though, so leaving it that way seemed worse than asking.

dominikh:
Thanks, I'll take a look eventually.

Eljees:
Correction to my comment of 9 August: I said I was marking this ready for review, but the conversion is refused — "Author has reached the open pull request limit". The repository's pull request limit counts non-draft PRs only, and #1730 is my one open non-draft PR here, so this one cannot leave draft while that is open. Apologies for the noise; I only found the real reason today.

So: #1730 first, and I'll flip this to ready as soon as it's merged or closed. If you'd rather look at this one instead, say so and I'll close #1730 myself.

